PR #4, opened 2019-04-26
added some design rules as discussed in the WG

alamers:


alamers:
as discussed in WG 9th of may

Issue #7, opened 2019-06-18
Open API spec fails linting


Jongmassey:
`             $ref: 'https://raw.githubusercontent.com/alamers/ICAR/convert-to-v3/Release%20Candidate%20Messages/JoinDataMilkSample.json'
`

from ln587 in https://github.com/adewg/ICAR/blob/master/Release%20Candidate%20Messages/icar-ade-openapispec.yaml

404 error reference to JoinDataMilkSample.json

PR #8, opened 2019-06-18
Fix open api spec linting errors

Jongmassey:
I linted the icar-ade-openapispec yaml using the speccy linter (http://speccy.io/) which found a non-resolving url reference (see https://github.com/adewg/ICAR/issues/7) and that the milking-visit-id parameter of many methods was missing a description. Description for this parameter was populated from icarMilkingVisitResource.json


cookeac:
Erwin's changes to the YAML file have removed these items back to the JSON schema files (now included by reference) and we have checked that these descriptions are in place in those files.

PR #32, opened 2019-07-24
 Add parentage to icarAnimalCore, add Movement events.

cookeac:


cookeac:
Hi Erwin,

1. Departure reasons - thanks for this. A bit of a mix of traditional departure reasons and health diagnosis. I can have a go at making the departureReason enumeration a superset of these and the reasons I have extracted from other systems.

2. Local I&R data - I have incorporated the common local I&R data from a number of English language movement systems I have access to. It would be helpful to get the Netherlands and Belgium I&R data fields (ideally with English explanations where possible) so we can see if any common fields can be introduced. I agree that anything that is truly "local" should be added when needed.
I recall one of the I&R fields was what we would call Coat Colour (or just "Color" in some systems). I propose we add that one as a string field to the Animal Core data (though it may also need to be in the event data if used for Departures).

3. Registration APIs - yes, I agree. I think you had started the process of defining these in the spreadsheet?

cookeac:
I have added coatColour to icarAnimalCoreResource.
I split icarMovementDepartureResource.departureReason into
- departureKind - the kind or type of departure (export, transfer, sale, show, slaughter, etc) and
- departureReason - the reason that animal was selected (e.g. health or production problems).
I appreciate this may make the mapping more complex, so open to feedback. It just seems there are two different things happening - the type of departure, and the decision behind it.

erwinspeybroeck:
> I have added coatColour to icarAnimalCoreResource.
> I split icarMovementDepartureResource.departureReason into
> 
> * departureKind - the kind or type of departure (export, transfer, sale, show, slaughter, etc) and
> * departureReason - the reason that animal was selected (e.g. health or production problems).
>   I appreciate this may make the mapping more complex, so open to feedback. It just seems there are two different things happening - the type of departure, and the decision behind it.

erwinspeybroeck:
Andrew,
coatColor - I think it might be difficult to get everybody on an ICAR standard color set ... 
Belgium - BlackAndWhite, RedAndWhite, BlueAndWhite, Grey, Red, White, Black, DarkBrown, LightBrown
Netherlands - BlackAndWhite, RedAndWhite, BlackPied, RedPied, BluePied, Brown, Grey, Red, White, Black, ThreeColored, PalePied, (then some colordrawings, rather than colors), WitrikRed, WitrikBlack, Blaarkop Red, Blaarkop Black, FlounderedRed, FlounderedBlack.

I agree on splitting departure reasons in 2 sets - is more in line with the deathReasons - let's see how the others react?

Regarding I&R:
In Belgium it is mandatory to report the "version number of the Sanitel card" in case of an arrival and departure.
In case of birth we have in Belgium and the Netherlands the following items:
"code name" : 1 = automatic, 2 = serial number, 3 = give a name (in next field), 4 = no name, 5 = automatic starting with an annual initial. 
Registry card (N/Y) : indicates whether the farmer is going to keep the animal and wants a registry card.
In Belgium in case of birth we need to provide two extra fields : 
RaceType (Dairy, Beef, Mixte) : this is used for suckler cows premiums 
RecentPurchaseMother (N/Y) : this is used when processing the birth and this field has a Y, then it is possible that the dam is not yet registered on the farm

Regarding the example url schema's : I will process your pull request, and I will add urls to the example url file.

kind regards,

Erwin

PR #38, opened 2019-08-15
Merge from ADEWG Master

cookeac:


cookeac:
Pull request the wrong direction; ignore.

PR #49, opened 2019-11-25
splitting up milking-visits API in milking-visits API and milking-visits-extended API

erwinspeybroeck:
Added also the units as documentation to the characteristics

### DIFF
--- a/Release Candidate Messages/exampleUrlScheme.json
+++ b/Release Candidate Messages/exampleUrlScheme.json
@@ -34,6 +34,125 @@
     }
   ],
   "paths": {
+    "/locations/{location-scheme}/{location-id}/milking-visits-extended": {
+      "get": {
+        "operationId": "get-milking-visits-extended",
+        "summary": "Get the extended data for milking visits",
+        "description": "# Purpose\nProvides data from visits of animals to a milking parlour\n",
+        "tags": [
+          "release-candidate-milk"
+        ],
+        "parameters": [
+          {
+            "name": "location-scheme",
+            "in": "path",
+            "description": "The scheme id for the location identifier.",
+            "required": true,
+            "schema": {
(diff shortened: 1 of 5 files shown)


cookeac:
I’ve held off merging this, because I thought we had agreed that as Characteristics and Quarters were already optional in icarMilkingVisitEventResource, we would not need to change the data structures, just create a separate API in Swagger and document that it was only expected to record basic milking visit and yield information?

Just checking this.

Thanks,
  Andrew.

erwinspeybroeck:
Andrew,

After discussion with Lely, we decided to make milking-visits the limited one and to create an extended version with the quarter info and the characteristics.
That’s why I did it like this. Seeing your thoughts we could maybe better decide that on ICAR level we could stick to a milking-visits with all information, and not create an extended version or limited version there. If some datahub like JoinData wants to create 2 seperate API’s based on the ICAR structure it would then be their choice. That’s maybe a better solution?

Kind regards,

Erwin


Van: Andrew Cooke <notifications@github.com>
Verzonden: dinsdag 26 november 2019 7:37
Aan: adewg/ICAR <ICAR@noreply.github.com>
CC: Erwin Speybroeck <erwin.speybroeck@crv4all.com>; Author <author@noreply.github.com>
Onderwerp: Re: [adewg/ICAR] splitting up milking-visits API in milking-visits API and milking-visits-extended API (#49)

I’ve held off merging this, because I thought we had agreed that as Characteristics and Quarters were already optional in icarMilkingVisitEventResource, we would not need to change the data structures, just create a separate API in Swagger and document that it was only expected to record basic milking visit and yield information?

Just checking this.

Thanks,
Andrew.

erwinspeybroeck:
I will create a new one - so this one will be closed

PR #67, opened 2020-02-12
Weight Event Resource and matching Collection

cookeac:
Added Weight Event (agreed fields only; more changes could be made). 
Edited icarResourceCollection so it no longer inherits from icarResource.


erwinspeybroeck:
Andrew, In issue 53 I also saw :
2. Add Operator attribute to capture person responsible for manual weighing. Discussed and accepted at last meeting.
3. Add TimeOffFeed attribute (hours) to capture time between animals moved from pasture or feed access to weighing (time to "empty out", hopefully standardising by reducing gut fill). 0 where animals are weighed in a feed environment. Discussed and accepted at last meeting.

You decided not to add them to the weight event? Or forgot them?

Erwin

cookeac:
When I did the first commit into my branch, we hadn't had the meeting and added those attributes which came from the Australian group of organisations (another 20 or so companies).
They have another couple of suggestions which I will pose at the meeting and then make a commit with all those changes.

Issue #73, opened 2020-02-14
Adding collection for all type of events resources


ahokkonen:
Resource collection messages should be added for all types of event resources (as now we have for icarWeightEventResource - icarWeightEventCollection). This could be assigned on me if no one working on that already.

examples:
_icarReproInseminationEventResource -> icarReproInseminationEventCollection
icarAnimalCoreResource -> icarAnimalCoreCollection
etc._

cookeac:
Yes, that was exactly the intent, and @ahokkonen it would be great if you could make a start on this.

PR #77, opened 2020-02-18
Added resource type collections

ahokkonen:
- added new collections for all main resource types
- renamed TestDayResource into TestDayResultEventResource
- fixed typo in icarWeightEventCollection (allof->allOf due case sensitivity)
- replaced responses in exampleUrlSchema for all endpoints to return typed collections instead of array of resources
- wrapped examples into type based collections

Resolves adewg/ICAR#73


cookeac:
Nice work, thank you

PR #101, opened 2020-05-14
Changes/fixes for better code generation

### LINKED ISSUE #100: Generated java code does not compile
When I generate java code with `java -jar openapi-generator-cli-4.2.3.jar generate -g jaxrs-resteasy -i ../src/main/resources/git-icar/url-schemes/exampleUrlScheme.json -o output` the resulting code does not compile. 

The Collections (i.e. IcarAnimalCoreCollection) extends IcarResourceCollection that has a `member` of type List<Object>, but i.e. has the same field `member` of type List<IcarAnimalCoreResource>. 

This does not compile
```
 .../target/generated-sources/openapi/src/main/java/org/openapitools/model/IcarAnimalCoreCollection.java:[54,15]
 name clash: setMember(java.util.List<org.openapitools.model.IcarAnimalCoreResource>) in org.openapitools.model.IcarAnimalCoreCollection 
and setMember(java.util.List<java.lang.Object>) in org.openapitools.model.IcarResourceCollection have the same erasure, yet neither overrides the other
```

Do I missed a parameter? Is it a problem of the OpenAPI-Generator? What tool do you use to generate compilable java code (> Java 8)?

ahokkonen:
- removed abstract "member" property from icarResourceCollection scheme
- removed empty inline schemes from icarMilkingDryOffEventResource and icarReproAbortionEventResource

#100

### DIFF
--- a/collections/icarResourceCollection.json
+++ b/collections/icarResourceCollection.json
@@ -49,10 +49,6 @@
           "description": "Link to the last page of the collection, if any. Link relation: last."
         }
       }
-    },
-    "member": {
-      "type": "array",
-      "items": {}
     }
   }
 }
--- a/resources/icarMilkingDryOffEventResource.json
+++ b/resources/icarMilkingDryOffEventResource.json
@@ -3,11 +3,6 @@
 
   "allOf": [{
       "$ref": "../resources/icarEventCoreResource.json"
-    },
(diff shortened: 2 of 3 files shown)


alamers:
I linked issue #100 to this pull request

PR #123, opened 2020-06-25
removed redoc since it does not work anymore with current spec

### LINKED ISSUE #72: Call for missing documentation
As we are approaching version 1.0, we should look at the documentation as well. This ticket is mostly a call for missing documentation: what do _you_ think we should add to the docs?

alamers:


alamers:
Already updated the wiki

alamers:
This pull request closes https://github.com/adewg/ICAR/issues/72

PR #124, opened 2020-06-30
Add test day resource

### LINKED ISSUE #110: Resource for listing test days
In our use case we want to deliver data to a milk recording organization (MRO) by retrieving the data from our system via an ICAR interface and converting the data into an ADIS file that can be sent to the MRO. To be able to send the data, the user needs to know:

1. When was the end of the previous test day that was executed (because some MROs require/offer the user to send all milk visits since the last test day)
2. What was the start/end date/time of the current test day

With that information, the relevant data can be collected through the ICAR milk visits API.

The data about the test days is available in our system, however, there is no resource type in the ICAR interface that allows us to transfer the values.

My proposal would be to provide a list of test day resources via the ICAR interface. As a follow-up improvement, the "TestDayResult" resource could even reference to one of that resources so they can easily be related.

MetroMarv:
First (extendable) draft of a test day resource.

Resolves #110

### DIFF
--- a/collections/icarTestDayCollection.json
+++ b/collections/icarTestDayCollection.json
@@ -1,0 +1,21 @@
+{
+  "description": "Represents a collection of test days. Based on icarResourceCollection to provide paging etc.",
+
+  "allOf": [{
+      "$ref": "icarResourceCollection.json"
+    },
+    {
+      "type": "object",
+
+      "properties": {
+        "member": {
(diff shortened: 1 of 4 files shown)


MetroMarv:
PR ist based on the Develop branch now.

cookeac:
Nice work.

PR #127, opened 2020-07-03
Add animal set resource

### LINKED ISSUE #88: Group assignments missing (feeding, milking, barn)
During the process of implementing a data exchange using the ICAR format we recognized that animals cannot be assigned to groups right now. In our use case animals can be assigned to specific groups to define how they are fed, milked and in which barn area they are located.

A possible integration of that into the ICAR standard could be:
- add a new resource "animal group" that has the attributes "type" as an enumerated value (e.g. "FEEDING", "MILKING", "BARN") and "name" which describes the group as a string value
- allow for CRUD on the new resource type
- add a new attribute "assigned groups" to the animal that consists of a list of "animal groups"

MetroMarv:
Resolves #88 

First minimal proposal for a animal set resource. We use the term "set" instead of "groups" here, because it's more generic.

### DIFF
--- a/collections/icarAnimalSetCollection.json
+++ b/collections/icarAnimalSetCollection.json
@@ -1,0 +1,21 @@
+{
+  "description": "Represents a collection of animal sets. Based on icarResourceCollection to provide paging etc.",
+
+  "allOf": [{
+      "$ref": "../collections/icarResourceCollection.json"
+    },
+    {
+      "type": "object",
+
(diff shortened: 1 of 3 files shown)


MetroMarv:
PR is based on Develop branch now

cookeac:
Fixed (hopefully) a conflict between test days and animal sets PRs in exampleURLScheme examples section.

PR #133, opened 2020-07-29
Added resources for conformation score event (#97)

### LINKED ISSUE #97: Conformation Scores
We have a need to transfer Conformation Scores (recorded by inspectors, following the ICAR guidelines http://www.icar.org/Guidelines/05-Conformation-Recording.pdf) for a project in New Zealand.

I propose to document the proposed messages in this Issue for review and approval, perhaps based on the work done on scores by the Integrity Systems group in Australia as well as the ICAR Guidelines.

ahokkonen:
Proposal for conformation score event.

- one score per trait
- score value can be decimal
- all possible traits (for dairy and dual purpose) combined in one enumeration
- automated/manual scoring methods allowed
- included "device" for referencing device information in case of automated scoring

TODO: examples to be added

Resolves  #97

### DIFF
--- a/collections/icarConformationScoreEventCollection.json
+++ b/collections/icarConformationScoreEventCollection.json
@@ -1,0 +1,21 @@
+{
+  "description": "Represents a collection of animal conformation scores. Based on icarResourceCollection to provide paging etc.",
+
+  "allOf": [{
+      "$ref": "../collections/icarResourceCollection.json"
+    },
+    {
+      "type": "object",
+      
+      "properties": {
+        "member": {
+          "type": "array",
+          "items": {
+            "$ref": "../resources/icarConformationScoreEventResource.json"
+          },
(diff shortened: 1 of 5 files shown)


cookeac:
This looks great. I particularly like that the list of possible traits are combined and species/purpose independent (because the species or purpose will be known from the animal/location). I will merge as is, but we should think about:
a. Whether "trait" might be confusing as there is also a "traitLabel" in the event base class (but I wouldn't suggest a major change, perhaps traitScored....).
b. Have we checked the other species conformation trait types (as per the same ICAR Guidelines document) to ensure we have a complete list (from memory, the cattle ones were effective for all species).

ahokkonen:
@cookeac 

I agree, we can change property name to "traitScored" or "conformationTrait" to avoid confusions.
I'll check traits list for including types for all species described in guidelines.

PR #141, opened 2020-08-21
Added animalSetJoin and animalSetLeave events

### LINKED ISSUE #139: Animal Sets (groups) membership change events
After implementing **icarAnimalSetResource** (issues #88) we thought it would be useful to have events that could record the date at which an animal joined or entered a set, and left/was removed from a set.

As with our other events, these should probably be animal-oriented (events based on the Animal ID) and reference the set.
For example: 
* **icarSetArrivalEventResource** and
* **icarSetDepartureEventResource**

ghost:
Closes #139

### DIFF
--- a/resources/icarAnimalSetJoinEventResource.json
+++ b/resources/icarAnimalSetJoinEventResource.json
@@ -1,0 +1,19 @@
+{
+  "description": "The animal set join event records animals joining animal set resources, i.e. additions of animals to a set.",
+  "required": [
+    "animalSetId",
+    "animal",
+    "eventDateTime"
+  ],
+  "allOf": [
+    {
+      "$ref": "../resources/icarEventCoreResource.json"
+    },
+    {
+      "animalSetId": {
+        "type": "string",
+        "description": "Unique identifier in the source system for the animal set to be joined."
+      }
+    }
+  ]
+}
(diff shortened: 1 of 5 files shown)


resources/icarAnimalSetJoinEventResource.json
```diff
@@ -0,0 +1,19 @@
...
+    "animalSetId",
+    "animal",
+    "eventDateTime"
+  ],
+  "allOf": [
+    {
+      "$ref": "../resources/icarEventCoreResource.json"
+    },
+    {
+      "animalSetId": {
```
cookeac, line 13: This could instead be a reference to an animal set if we wished. A reference must provide an ID, but may optionally also provide the URL to the object, which might be handy. Not essential however.

resources/icarAnimalSetLeaveEventResource.json
```diff
@@ -0,0 +1,19 @@
...
+    "animalSetId",
+    "animal",
+    "eventDateTime"
+  ],
+  "allOf": [
+    {
+      "$ref": "../resources/icarEventCoreResource.json"
+    },
+    {
+      "animalSetId": {
```
cookeac, line 13: This could instead be a reference to an animal set if we wished. A reference must provide an ID, but may optionally also provide the URL to the object, which might be handy. Not essential however.

cookeac (approved):

PR #149, opened 2020-08-31
Add Animal health status

### LINKED ISSUE #148: Add Animal Health Status to Animal Resource
In issue #111 we discussed reproduction and lactation status values. In addition to this, a health status of an animal would also be very helpful. As possible values I'd like to propose:

- Healthy
- Suspicious
- Ill
- In Treatment
- To Be Culled

On the animal, this would be an additional field defining the current status. In addition to this, a Health Check Event could be added so that an observation of a changed health status can be filed.

I think, the Health Check Event could be a little bit more complicated because you might need diagnosis codes etc., however, if everyone is OK with the status field at the animal I could add the health status field already for the 1.1 release.

ghost:
Closes #148

### DIFF
--- a/collections/icarAnimalSetJoinEventCollection.json
+++ b/collections/icarAnimalSetJoinEventCollection.json
@@ -1,0 +1,20 @@
+{
+  "description": "Represents a collection of animal set join events. Based on icarResourceCollection to provide paging etc.",
+  "allOf": [
+    {
+      "$ref": "../collections/icarResourceCollection.json"
+    },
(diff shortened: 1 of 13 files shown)


ahokkonen (approved):

cookeac:
We decided on 06 May 2021 to close this PR and address again in the next release, as there is more work to do:

- Deciding which components are truly health status or other culling/removal decision
- Defining an event (or events) that alter this status.

PR #151, opened 2020-09-11
Add observation events for animal status - lactation and reproduction status

### LINKED ISSUE #111: Reproduction and lactation status values
For data exchange with a milk recording organization we need to be able to determine the current status of animals to report some data about them. In the current API there is no information about the animal status. I could imagine that one could derive the animal status from looking at the last animal actions, however, this seems quite cumbersome.

I'd propose to add an additional status to the AnimalCoreResource with enum values that reflect the current status. Possible enum values:

- other
- fresh
- early
- abort
- ready
- open
- inseminated
- pregnant
- pregnant2
- dry
- lead
- tbcull

cookeac:
This PR addresses the outstanding issue and closes #111


cookeac:
Merging so we can finalise 1.1

PR #169, opened 2020-10-16
Correct icarMedicineResource syntax

### LINKED ISSUE #167: Missing reference in icarMedicineResource -resource type
@cookeac @erwinspeybroeck @alamers 

Seems like "registeredID" property in "[icarMedicineResource](https://github.com/adewg/ICAR/blob/eaf265770971f56db76aa9518d4d41dc20ec0297/resources/icarMedicineResource.json)" -type is pointing into non-existing type "iscMedicineRegistrationIdentifierType.json". Something left unposted into the branch?

Noticed that during code generation for the new ADE v1.1 version

The reason why this had passed JSON validation is that MedicineResource is not part of example url scheme.
If running something like "spectral lint **/*.json" in root there are 2 issues with icarMedicineResource:

-   6:21  error  invalid-ref  ENOENT: no such file or directory, open 'icarResource.json'
-  20:29  error  invalid-ref  ENOENT: no such file or directory, open 'iscMedicineRegistrationIdentifierType.json'

cookeac:
- correct reference to **icarResource.json**
- correct *registeredID* to reference **icarMedicineIdentifierType.json**

Fixes #167

### DIFF
--- a/resources/icarMedicineResource.json
+++ b/resources/icarMedicineResource.json
@@ -3,7 +3,7 @@
     
     "allOf": [
         {
-            "$ref": "icarResource.json"
+            "$ref": "../resources/icarResource.json"
         },
         {
             "type": "object",
--- a/resources/icarMedicineResource.json
+++ b/resources/icarMedicineResource.json
@@ -3,13 +3,13 @@
     
     "allOf": [
         {
-            "$ref": "../types/icarResource.json"
+            "$ref": "icarResource.json"
         },
         {
             "type": "object",
             "properties": {
                 "name": {
-                    "type": "string",
+                    "type": "string", 
                     "description": "Name of the medicine or remedy given for this treatment"
                 },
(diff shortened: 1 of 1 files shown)


resources/icarMedicineResource.json
```diff
@@ -3,21 +3,21 @@
     
     "allOf": [
         {
-            "$ref": "../types/icarResource.json"
+            "$ref": "icarResource.json"
```
ahokkonen: @cookeac I think this should point into "../resources/icarResource.json" as we agreed on using root relative paths

ahokkonen (changes requested):

resources/icarMedicineResource.json
cookeac: Agreed.

ahokkonen (approved):

PR #177, opened 2020-12-03
Fixing probable typo in test-day-results Milk Characteristics weight

lkoivuka:
## Background
The description for milking characteristics doesn't seem correct. 
`/locations/{location-scheme}/{location-id}/test-day-results` has an array of milk characteristics, and within that array it seems like there's a milk weight, but it's described as animal weight. The unit is KGM as well, which I assume is Kilograms of milk, but am not familiar with that as an actual unit.

## Side note
There's additionally `milkWeight24Hours` in the test day as well, so I'm assuming the weight in the characteristics array is for the milking being tested.

## Proposal
|WEIGHT|Weight of animal|Kg|KGM -> |WEIGHT|Weight of milk|Kg|KGM

### DIFF
--- a/types/icarMilkCharacteristicsType.json
+++ b/types/icarMilkCharacteristicsType.json
@@ -11,7 +11,7 @@
   "properties": {
     "characteristic": {
       "type": "string",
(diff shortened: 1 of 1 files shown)


erwinspeybroeck:
Lars,

this characteristic is indeed the animal weight. A robot can weigh the animal when being milked. It might be considered as a misfit because it is not a charactertic of the milk, but in fact could be considered as a characteristic of the milking. 

Erwin

lkoivuka:
Ok, thanks for clarifying! I assumed it would be outside milk characteristics, but am fine knowing that it is indeed the weight of the animal, not the milk.

PR #181, opened 2021-01-11
Fixed reference in icarDailyFeedReportCollection

ahokkonen:
- fixed reference to base collection resource


erwinspeybroeck:
It seems I overlooked that one ...

Issue #180, opened 2021-01-04
Rank field in icarReproInseminationEventResource event


ahokkonen:
During standard implementation we've noticed that "Rank" field is missing from insemination event message.

Rank (the rank of intervention of each AI carried out within the same reproductive cycle) is the one field defined as recommended in ICAR general [rules ](https://www.icar.org/Guidelines/01-General-Rules.pdf)(2.4.4.10)

Technically this field could be calculated on server side, but in some cases there might be a need for providing it from the client.
Any obstacles for adding this new field to the icarReproInseminationEventResource message?

ahokkonen:
Rank field is now part of icarReproInseminationEventResource message after my PR was merged.
For most cases Rank value is increased by 1 with each new AI during same reproductive cycle, **also in case of double insemination**.

cookeac:
Fixed in #186

Issue #179, opened 2020-12-17
endpoint /animalSets uses inconsistent syntax


alamers:
Shouldn't `/locations/{location-scheme}/{location-id}/animalSets` be named `/locations/{location-scheme}/{location-id}/animal-sets`?

ahokkonen:
I agree with that proposal. I don't think this correction will be considered as breaking change, or? :)

cookeac:
I also agree with this proposal, and I note that the exampleURLScheme.json file is effectively INFORMATIVE rather than NORMATIVE - it provides an example or recommended scheme (in contrast to data elements where a breaking change could break software). 

However, because some implementers may have built on this example, I recommend that we document this change clearly when publishing the release with this change.

cookeac:
Fixed in #186

PR #186, opened 2021-02-07
Added new field to insemination resource + fixed "animal-sets" endpoint name in example url scheme

ahokkonen:
- added field "Rank" to "icarReproInseminationEventResource" resource (#180)
- "animalSets" endpoint name is fixed to "animal-sets" (#179)

### DIFF
--- a/resources/icarReproInseminationEventResource.json
+++ b/resources/icarReproInseminationEventResource.json
@@ -12,6 +12,10 @@
       ],
 
       "properties": {
+        "rank": {
+          "type": "integer",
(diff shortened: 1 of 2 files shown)


cookeac (approved):

alamers:
When making changes wrt issue 130 (filter parameters), I noticed that this resource has a badly named 'member' field:
```
 "member": [
    {
      "member": [
        {
          "scheme": "eu.animalId",
          "id": "276000312312345"
        },
```

The second `member` should probably be called `animal` to be consistent with animal-set-joins etc.

PR #197, opened 2021-02-25
Add sourceId to icarMetaDataType

### LINKED ISSUE #153: Unique resource identifiers when sending new data to a server
This issue is a placeholder to be elaborated. 
It covers the question of how objects should be identified when sent to a server by a client - using UUIDs, client issued identifiers, assuming the server will issue identifiers, etc.

cookeac:
resolves #153


erwinspeybroeck (commented):
Andrew, I agree. This is how we are implementing it at the moment with Lely.

Issue #194, opened 2021-02-17
Feed Storage (e.g. Silo) in ICAR?


ghost:
When having a connection between herd management systems, exchanging the feed information (as covered in #93) might not be sufficient. However, is this sort of information covered in ICAR? I believe this is more "process related information", however, it might still be important for the users to exchange that kind of data.

What I'd like to have is a new resource (e.g. feed storage to be generic) that would contain the following info:
* name (string)
* type (string or enum)
* contained feed (reference to feed id)
* capacity (number, connected to unit below)
* quantity unit (enum)
* contained feed amount (number, connected to unit above)

Any opinions on this?

ghost:
I created a first example. Here, the feedStorageResource "inherits" all information from the deviceResource. Would be happy to get feedback.

cookeac:
Resolved by #198

PR #198, opened 2021-02-26
Feed storages

ghost:
First proposal for #194

### DIFF
--- a/resources/icarFeedStorageResource.json
+++ b/resources/icarFeedStorageResource.json
@@ -1,0 +1,32 @@
+{
+  "description": "Describes the feed storage device on a certain location.",
+
+  "type": "object",
+
+  "required": [
+    "id"
+  ],
+
+  "allOf": [{
+    "$ref": "icarDeviceResource.json"
+    },
+    {
+      "type": "object",
+
+      "properties": {
+        "feedId": {
+          "type": "string",
+          "description": "Unique identifier of the feed that is stored in this device."
+        },
+        "capacity": {
+          "$ref": "../types/icarFeedQuantityType.json",
+          "description": "The amount of feed that can be stored in this device."
+        },
+        "availableFeed": {
+          "$ref": "../types/icarFeedQuantityType.json",
+          "description": "The amount of feed that is currently stored in this device."
+        }
+      }
+    }
+  ]
+}
--- a/resources/icarFeedStorageResource.json
+++ b/resources/icarFeedStorageResource.json
@@ -22,9 +22,9 @@
           "$ref": "../types/icarFeedQuantityType.json",
           "description": "The amount of feed that can be stored in this device."
         },
-        "availableFeed": {
+        "quantityAvailable": {
           "$ref": "../types/icarFeedQuantityType.json",
-          "description": "The amount of feed that is currently stored in this device."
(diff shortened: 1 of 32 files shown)


thomasd-gea:
I created a first example. Here, the feedStorageResource "inherits" all information from the deviceResource. Would be happy to get feedback.

resources/icarFeedStorageResource.json
```diff
@@ -0,0 +1,32 @@
...
+      "properties": {
+        "feedId": {
+          "type": "string",
+          "description": "Unique identifier of the feed that is stored in this device."
+        },
+        "capacity": {
+          "$ref": "../types/icarFeedQuantityType.json",
+          "description": "The amount of feed that can be stored in this device."
+        },
+        "availableFeed": {
```
cookeac: I did wonder if availableFeed should be quantityAvailable, to avoid any confusion that it might be a feed ID.

resources/icarFeedStorageResource.json
```diff
@@ -0,0 +1,32 @@
...
+  ],
+
+  "allOf": [{
+    "$ref": "icarDeviceResource.json"
+    },
+    {
+      "type": "object",
+
+      "properties": {
+        "feedId": {
```
cookeac, line 17: Is there any possibility that a feed storage resource might store more than one feed? Or might possibly store several feeds, and the feedId is only the current feed (or last feed if empty)?

resources/icarFeedStorageResource.json
cookeac: Should there be units associated with a feed storage resource and feed type? These units would presumably be the same for both capactity and quantity available. For example, see enums/uncefactMassUnitsType.json

ghost: Yes, "quantityAvailable" is much better :) Actually, I was struggling finding a good name but didn't find any. I'll change that.

As for your second comment: I don't fully understand. The "icarFeedQuantityType" already contains a unit. I agree that different units for capacity and quantityAvailable does not make sense, however, I wanted to reuse the type that we have there. If you prefer I can change the fields to reference only numbers and add a unit to the feedStorage directly.

resources/icarFeedStorageResource.json
ghost: My idea was that each feedStorage device can only contain one feed at a specific point in time. The question would be, whether "rations" or "mixes" can be contained. For both I'd propose that the user creates "feed" entries that just gives those an id and a name and then it can be referenced in a storage device easily. Of course, at any time the user can decide to change the contents of a feed storage by just switching to another feed (and adapting capacity, quantityAvailable and possibly unit accordingly). Is that sufficient?

resources/icarFeedStorageResource.json
ghost: I've updated the field name.

cookeac: Great. And you are right - I was thinking about making shared units (for both *capacity* and *quantityAvailable*) explicit - but on reflection I think it is better to reuse the existing **icarFeedQuantityType**.

resources/icarFeedStorageResource.json
cookeac: Yes, that makes sense to me.

cookeac:
Resolves #194

PR #201, opened 2021-03-01
Correct example URL scheme

cookeac:
Correct typo in scheme from #198


erwinspeybroeck:
Noticed it too - correction ok for me

PR #208, opened 2021-03-25
Add Calving Ease to fix #185

### LINKED ISSUE #185: Define enumeration for icarReproParturitionEventResource - calvingEase
Currently the **calvingEase** attribute of *icarReproParturitionEventResource.json* is a string field with no definition.
There would be benefit in a consistent enumeration or score.

Splitting this issue out from the remaining discussion in #183 so that it can be addressed specifically. This will allow further research into the other topics in that issue.

**Research: Would this cause a breaking change for current implementations?**

cookeac:
Add icarReproCalvingEaseType.json with 5 values corresponding to the Interbeef codes 1-5.
Fixes #185


ahokkonen (approved):
Looks perfect :)

PR #217, opened 2021-05-20
Example URL scheme with POST for Health

### LINKED ISSUE #154: Semantics for POST method
As part of our work on URL schemes and recommended common implementations, this issue covers using HTTP POST to send new resources from a client to a service. This issue is defined to allow discussion. I've made a start below:

## Goals
- Provide a standard approach that can be implemented by any server that allows clients to add new resources
- Work for any object collection where the server supports addition (for instance, animals, animal sets, events)
- Align with best practices for HTTP, REST, and JSON APIs
- Allow the posting of single resources and collections of resources
- Provides a standardised way of handling errors.

There are several parts to this issue:
1. How to determine if a collection implemented on the server supports adding resources
2. The method and URL path to use when adding new resources
3. Adding single resources or arrays/collections of resources
4. How to uniquely identify resources being added to the server
5. Determining status - success or errors
6. Return results when resources are successfully added
7. Return results when the server is processing new resources asynchronously
8. Returning error results.

## How to determine if a collection on a server supports adding resources
There are three options:
- Read the human-readable documentation or talk to an administrator
- Retrieve an icarResourceCollectionReference that provides a reference to the collection, and parse the operations array to see if the "POST" method is supported - this could be through a capabilities or endpoint enumerating API.
- Have some other standard capabilities API

## The method and URL path to use when adding new resources
1. The method should be HTTP POST as per [RFC-2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.5)
2. The URL should be the URL which one would GET to get a collection of resources of the specified type. For instance, the URL of the animals collection for icarAnimalCoreResource objects.

## Adding single resources or arrays/collections of resources
There might be several options:
- Only support posting individual items (likely inefficient for adding bulk data)
- Define a special purpose array/collection for posting
- Use an **icarResourceCollection**, setting the *memberType* and putting resources in the *members[]* array. The *view* properties would be unused.

## How to uniquely identify resources being added to the server
A discussion area in itself - broken out into issue #153 

## Determining status - success or errors
The recommended method for REST web services is to use [HTTP status codes ](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10).  For instance,
- **201** Created
- **202** Accepted
- **4xx** for client errors (for instance, **409** Conflict when a resource with the same unique ID already exists)
Substantial work on this has been done by JSON-API (https://jsonapi.org/format/#crud-creating-responses).

## Return results when resources are successfully added
Based on the RFC-2616 specifications and JSON-API, it would seem that when new resources are added successfully, the server should:
1. Return HTTP status 201 Created
2. Set the HTTP Location header to the location of the new resource(s)
3. Return the newly added resources (the same data sent, possibly with IDs filled in and other values generated) in the body.

 ## Return results when the server is processing new resources asynchronously
Based on the RFC-2616 specifications and JSON-API, it would seem that when the server is still processing the submitted, resources (and would otherwise time out), it should:
1. Return the HTTP status 202 Accepted
2. Return **something** in the body that indicates status and how to check status or the result. Some examples are given here: https://restfulapi.net/http-status-202-accepted/ (we will need to decided what we want to do).

##  Returning error results
The usual recommended approach is to return an HTTP 4xx client error status, and then return an error or problem response in the body.
- [RFC-7807](https://tools.ietf.org/html/rfc7807) defines a form of problem response that can be extended.
- JSON-API defines an [Errors object](https://jsonapi.org/format/#errors)
We will need to decide the representation we want to use AND whether a server should always fail when an error (rather than a warning) is encountered, or whether it should attempt to process other objects in the request.

cookeac:
This is a first experiment with a separate URL scheme for the tag ADE-1.x-health.
It covers health diagnosis, treatment, and treatment programme.
Methods supported are
* GET (collection)
* POST (to collection - single event)
* POST (batches)

This implements SOME of the thinking in #154 but is not yet a fix. More work to be done on Error handling, for example.


cookeac:
Created this for discussion/comment at the meeting. It can be removed after if on the wrong track.

url-schemes/healthURLScheme.json
```diff
@@ -0,0 +1,611 @@
...
+            "$ref": "#/components/parameters/location-id"
+          }
+        ],
+        "requestBody": {
+          "required": true,
+          "description": "The collection of health treatment events to create. \nA client MAY fill in resource *Id*s with a client-generated UUID and the server MAY use these *Id*s.\nIf the server does not use the client-specified *Id* field it shall issue its own *Id*s for the resources.\nA client SHALL ensure that the *meta.source* and *meta.sourceId* fields are filled by the client.\nAlthough a collection is used, servers SHALL IGNORE any pagination-related attributes.\n",
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/icarTreatmentEventCollection"
```
ahokkonen: Do we want to use collection type for posting resources as a batch? Could it by just an array of resources (aka array of icarTreatmentEventResourcce)?

url-schemes/healthURLScheme.json
```diff
@@ -0,0 +1,611 @@
...
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "Successful. The response contains a copy of the collection, as modifed by the server.",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/icarDiagnosisEventCollection"
```
ahokkonen: This approach only supports "all or nothing" posting - if all ok collection of created resources are posted back if not - list of errors receved. What if we need to support partial success? We need to get statuses on each resource - success or failure with relevant errors.

ahokkonen (commented):
Left some review comments.

Our main concern is that this approach would fit perfectly atomic operations, but in our solution, we need to support non-atomic scenarios where party failure/success is allowed. For that we need a mechanism to determinate statuses of the sent resources back from the service response. 
So, it is not enough to just have a same resource back in response, but also a status and possible error array for failed resources.

cookeac:
We discussed both of those previously at our meeting. This does sound like a fairly specific case that is not well supported by widely used standards such as problem details that are indeed atomic. I'd be interested in the HTTP status coffee that you would propose to return when accepting some data but rejecting others. What would this look like to be REST compliant?



Andrew Cooke, Managing Director, Rezare Systems

m: +64-27-5570822<tel:+64275570822> p: +64-7-8570822<tel:+6478570822>


ahokkonen:
@cookeac 
that is true - our case and requirements are a bit out of specs. For https statuses and response we where thinking about the following approach:

1. Batch operation must return a success status code (200) only if processing of all resources is succeeded.
2. If any single resource handling fails the whole batch operation must return the failure status code (4xx)
3. Service response must contain processing information about every resource from batch request including successed and failed. For successed we need to get back server updated information (server generated identifier, for example) and for failed we need to get an information about the failure.

**Example for the batch response for one successed and one failed resource handling response**
```json
[
    {
        "id": null,
        "meta": {
            "source": "XXXX",
        },
        "errors": [
            {
                "id": null,
                "status": 400,
                "code": null,
                "title": null,
                "detail": "Required field cannot be null: Meta.SourceId",
                "meta": null
            }
        ]
    },
    {
        "id": "5",
        "meta": {
            "source": "XXXX",
            "sourceId": "352178635217635126783",
            "modified": "2021-06-15T11:12:10.7779154Z",
            "creator": "iDDEN.Stub"
        },
        "errors": null
    }
]
```

url-schemes/healthURLScheme.json
cookeac: It could. I was trying to avoid creating another data structure for the return and therefore just reusing the collection. However, a collection is also decorated with next and previous links which have no meaning and would be unsupported when POSTing data.

ghost:
As discussed in our meeting, it would be nice if we could add "warnings" (similar to errors) so that the receiving system can add those to entities that were accepted, but still something important needs to be reported back. In general, this could also be interesting for a single post as well.

cookeac:
We should revisit batches. Maybe a REST compliant single item POST and a /batches/ post that does not attempt to be compliant with REST norms but instead returns a special purpose result that deals with the identifiers, errors and warnings as discussed. A sort of morph of problem details and success. It should support multi-value error returns.

url-schemes/healthURLScheme.json
cookeac: Changed to use arrays for POST. Left the single POST as it is.

cookeac:
Changed **/batches/ POST** to return an array of **icarBatchResult** instead (in case of both 200 and error). This will contain:
* id - issued id
* meta - incl source/sourceId
* errors (array of icarErrorResource)
* warnings (array of icarErrorResource that represents warning messages)

url-schemes/healthURLScheme.json
cookeac: Adjusted as per your recommendation. See overall comment below.

resources/icarBatchResult.json
```diff
@@ -0,0 +1,30 @@
+{
+    "description": "Returned by a batch POST event to return identity (meta), errors, and warnings for a resource.",
+    "type": "object",
+    "properties": {
+        "id": {
+            "type": "string",
+            "description": "Unique identifier created in the system for this event. SHOULD be a UUID."
```
ahokkonen, line 7: Is this field used for identifier created for the event by the receiver side (server)?

resources/icarBatchResult.json
```diff
@@ -0,0 +1,30 @@
...
+        },
+        "meta": {
+            "$ref": "../types/icarMetaDataType.json",
+            "description": "Metadata for the posted resource. Allows specification of the source, source Id to synchronise data."
+        },        
+        "errors": {
```
ahokkonen: could errors and warnings be combines under one field ("norifications") and icarErrorResource (maybe then resource name should be different) have an additional field "severity" to identify the severity level of the message? with that approach we could pass errors, warning and infos withint the same set of resources?

ahokkonen (commented):
added some comments/questions

resources/icarBatchResult.json
cookeac: Yes, that's correct.

resources/icarBatchResult.json
cookeac: I wondered that too - I separated them because then the client can quickly check to see if an array is empty or not rather than having to look at each item for severity, but I'm open to doing it the other way as well.

ahokkonen: I do not have any strong feelings about it :) both solutions would be fine in my opinion. 
If we'll decide to go with the combined list then we'll need to define enumeration for severity levels - Info, Warning, Error (maybe Critical?) - where "Error" is an indication for event failure and other are just an additional notification for successed processing.

cookeac: Renamed icarErrorResource to icarBatchResultMessage, and added icarBatchResultSeverityType enum. Removed errors[] and warnings[] arrays and replaced with messages[].

alamers: It seems the rename was not fully done yet? When I run `java -jar ~/local/openapi-generator-cli-5.2.0.jar generate -g jaxrs-resteasy -i https://raw.githubusercontent.com/cookeac/ICAR/master/url-schemes/healthURLScheme.json` I get a lot of messages stating:
`Caused by: java.io.FileNotFoundException: https://raw.githubusercontent.com/cookeac/ICAR/master/resources/icarErrorResource.json`

alamers: (`exampleUrlScheme.json` generates fine)

PR #222, opened 2021-07-05
MilkingRemarks

AndreasSchultzGEA:
closes issue 220


cookeac (approved):
Looks good. I'm not sure if the enumeration has the full list of milking remarks needed, but it is a starting point and can be extended.

PR #226, opened 2021-07-19
added positions in the treatment and diagnosis resource

erwinspeybroeck:


types/icarPositionType.json
```diff
@@ -0,0 +1,26 @@
+{
+  "description": "The possible positions for treatment or diagnosis",
+
+  "type": "object",
+
+  "properties": {
+    
+    "position": {
```
ahokkonen, line 8: Maybe this should be extracted to the separate enum file and values adjusted according to the naming convensions (aka LEGS_FRONT_LEFT -> FrontLeftLeg)?

ahokkonen (commented):
Left some comments about position enumeration

types/icarPositionType.json
cookeac: I agree, this would make it easier to extend in future.

cookeac: Agreed at the meeting today that I will make this change and then make a PR with it. So I will merge this PR and then make the change.

PR #232, opened 2021-07-30
Example Health URL scheme with GET, single POST and batch POSTs

### LINKED ISSUE #154: Semantics for POST method

cookeac:
Resolves #154 
* Adds healthURLScheme.json (OpenAPI example for animal health events)
* Adds icarBatchResult.json - results for batch POSTs
* Replaces icarErrorResource.json with icarErrorWarningMessageResource.json which is used for errors but also errors and warnings in batch results.
* Adds icarBatchResultSeverityType.json to provide an enumeration for error, warning, and informational messages.

In usage:
* GET methods function as they do in exampleURLScheme.json
* Single POST methods take a single resource and return the updated resource if successful, else an array of error/warnings.
* Batch POST methods take an array of resources and return a matching array of batch results, with the original metadata for matching, the updated resource IDs (where added successfully) and/or an array of errors and warnings for each resource.


resources/icarErrorWarningMessageResource.json
```diff
@@ -0,0 +1,35 @@
+{
+  "description": "An RFC7807 compliant problem response for JSON APIs.",
```
ahokkonen: Could "icarResponseMessageResource"/"icarProblemResponseMessageResource"  be a better name for that resource? As we also have "Information" severity level defined for message type.

ahokkonen (commented):
Added one comment

resources/icarErrorWarningMessageResource.json
cookeac: Excellent suggestion, thank you. I struggled with the naming of that resource. Renamed to icarResponseMessageResource.

ahokkonen (approved):
Looks perfect now 👍🏻 Great work!

AndreasSchultzGEA (approved):
Hi, I had no closer look at the UrlScheme, but the other parts look fine.

Issue #191, opened 2021-02-12
adding 2 Types in ReproParturitionEventResource


dirbSEGES:
Hi 

We will make local additions in ReproParturitionEventResource 

"progenyBirthsizeType"  with following selections: 

Little
Below average
Above average
Big

and 

 "progenyStateofbirthType" with following selections:  

Defective calf
Live calf
Live calf, born prematurely 
Died within 1 day
Died after 1 day
Stillborn
Abortion
Culled at birth

To accept a parturition theese input will required from our side. 

Maybe you're considering being added to ADE?

ahokkonen:
@dirbSEGES 

This is also one of the things we are missing compared to previous exchange standard using in our systems (NordicNDX).

**BirthSize** and **BirthStatus** are the parameters what needs to be passed on progeny level, but for progeny we have an array of icalAnimalCoreResource  -resources defined in a message. One solution would be to extend icalAnimalCoreResource with new fields, but I'm not sure is it a proper way of solving that.

Another way is to define new type aka icarAnimalBirthResource/Type (inherited from icalAnimalCoreResource ) and have it as progeny message. Then we can add required fields there

dirbSEGES:
@ahokkonen 

both solutions are usable for us. We need this info to create the parturition correct.

tpekeler:
We would like to have the values above added to the standard.

In case of birthSize we would add another value "average" for countres where it is needed.

For the birthState we think it would be good to have only one set of values and everyone should pick the values which ar needed.
This could be done by providing only translations in the local language for the desired values. the client could then filter the values based on the translation tables.

Would this be a solution also to other EnumValues in the ADE standard?

cvigorsICBF:
We would collect multiple events/types at birth:

- Malpresentation:  no, backwards, head down, one leg down, both legs down, other
- Birth Size: extra small, small, medium, large, very large
- Birth Vigour: very lazy/weak, lazy, average, vigorous, very vigorous
- Birth Weight: kgs
- Chest Girth: cm
- Shoulder height: cm

Each value would be recorded against the calf

erwinspeybroeck:
At birth we collect:
calvingEaseScore : EASY, NORMAL, HEAVY, CAESAREAN, SAWED_OFF, OTHER_HELP
calfMeatScore : LOW_MUSCULATED, NORMALLY_MUSCULATED, EXTRA_MUSCULATED, DOUBLE_MUSCLED
[...]

PR #243, opened 2021-08-24
Changes for parturition event with new progeny details

ahokkonen:
Resolves adewg/ICAR#191

Added new types for progeny details and enumeration parameters for progeny birth statuses.
**Note**! _This is a draft PR - work is still in progress and will be progressed after first reviews/comments._

### DIFF
--- a/enums/icarParturitionBirthSizeType.json
+++ b/enums/icarParturitionBirthSizeType.json
@@ -1,0 +1,13 @@
+{
+  "description": "Enumeration for the widely used progeny birth sizes.",
+
+  "type": "string",
+
+  "enum": [
+    "ExtraSmall",
+    "Small",
+    "Average",
+    "Large",
+    "ExtraLarge"
+  ]
+}
--- a/enums/icarParturitionBirthStatusType.json
+++ b/enums/icarParturitionBirthStatusType.json
@@ -1,0 +1,15 @@
+{
+    "description": "Enumeration for the widely used progeny birth statuses.",
+  
+    "type": "string",
+  
+    "enum": [
+      "Alive",
+      "Stillborn",
+      "Aborted",
+      "DiedBeforeOfficialRegistrationDate",
+      "DiedAfterOfficialRegistrationDate",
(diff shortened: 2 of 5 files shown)


erwinspeybroeck (approved):
Looks good to me. I wonder if it would not be better to leave out the progeny array at all - but probably needed for backwards compatibility I guess - for version 2.0 ....

ahokkonen:
@erwinspeybroeck as far as I remember we agreed on leaving the original array, but marking it as 'deprecated'. And we can then remove it completely in 2.x

dirbSEGES:
Hi, what about the birthWeigtScore?

cookeac:
This is in progress. We discussed at the last meeting how these "scores" could be handled as enumerations, but I also remind you that you can always extend the schema for days that only your country or application uses.


ahokkonen:
@erwinspeybroeck @dirbSEGES

New field is added for "birthSize" using enumeration for common codes.

@cookeac I'm still a bit unsure we we want to use Resource based type for icarParturitionProgenyDetailsType, any reason for having it like that?

ahokkonen:
One thing in addition - should also sire information be added for the progeny details or for the parturition event?

erwinspeybroeck:
looking from our perspective I would say no sire information to be added there. We always derive the sire based on reproduction events

dirbSEGES:
The same from our perspective

ahokkonen:
Added a few changes - I guess having "DiedBeforeTaggingDate" instead of a "DiedBeforeOfficialRegistrationDate" is a more correct term. What you think @erwinspeybroeck @dirbSEGES @cookeac

dirbSEGES:
Ok, from our side.

erwinspeybroeck:
As long as the definition of what is meant is right - in fact it means the time the farmer has to officially tag the animals, I guess

ahokkonen:
PR opened for the official review
@cookeac @erwinspeybroeck @alamers

cookeac (approved):

PR #252, opened 2021-09-29
CoatColorDetails in IcarAnimalCoreResource

### LINKED ISSUE #240: Alternative Type for Coat Color
As mentioned in our meeting on 12.08.2021:

The coatColor property of the icarAnimalCoreResource.json states:
`Colour of the animal's coat, using the conventions for that breed.`

That is not very helpful as it seems to imply that the primary breed scheme/identifier should have an influence on this. As discussed, there would be two ways on how we can improve this:

1. change the coatColor to contain an enum value (have to make a list of all possible coat colors and each receiving party has to maintain a mapping from possible coat colors to its own list of allowed values)
2. use scheme/identifier instead to allow for different enumerations per country/organization/... (but we could still provide a default scheme, if we want to)

Any opinions which way to go?

We should also decide if this should be solved in a 1.x release, because then we have to add an additional field for the new coat color to be backwards-compatible and might have to rename it for a 2.x version.

AndreasSchultzGEA:


erwinspeybroeck (approved):
Maybe we should add some example schemes in the Wiki. This has also been done for a number of other schemes --> https://github.com/adewg/ICAR/wiki/Schemes

AndreasSchultzGEA:
added Animals.CoatColor to https://github.com/adewg/ICAR/wiki/Schemes
Pls. have a look

resources/icarAnimalCoreResource.json
```diff
@@ -49,6 +49,10 @@
           "type": "string",
           "description": "Colour of the animal's coat, using the conventions for that breed."
         },
+        "coatColorIdentifier": {
+          "$ref": "../types/icarCoatColorIdentifierType.json",
+          "description": "Colour of the animal's coat, using the conventions for that specie."
```
cookeac: I'd rather the description said "Colour of the animal's coat using a national or breed-defined scheme and identifier combination."

cookeac (changes requested):

cookeac (approved):

PR #262, opened 2021-11-17
Erwinspeybroeck

erwinspeybroeck:
Added API for type-classification and changed the conformation API to reuse the conformation part


resources/icarConformationScoreEventResource.json
```diff
@@ -1,35 +1,18 @@
...
-  }
\ No newline at end of file
+	"allOf": [{
+			"$ref": "../resources/icarEventCoreResource.json"
+		},
+		{
+			"type": "object",
+
+			"properties": {
+				"conformationscore": {
```
cookeac: This is a breaking change, but if the icarConformationScoreType was incorporated using "allOf" it could be a non-breaking change I think.

PR #263, opened 2021-11-17
Draft generic api

gra-moore:
This is a draft outline of the generic data sharing specification and also plays with ideas around semantic typing, identifiers on resource level.


gra-moore:
newer PR incoming.

PR #269, opened 2021-12-07
Type classification

### LINKED ISSUE #247: type classification
We have an API description for a conformation score event.
Herdbooks are doing type classifications = scoring a number of linear traits and composite traits (thes might be calculated from linear traits) at once for a number of animal on a location

So we might consider an extra API : type-classifications. This would then comprise in one event an array of linear and an array of composite traits.

erwinspeybroeck:
I solved the breaking change - needs to be checked if code generation is going ok


url-schemes/performanceURLScheme.json
```diff
@@ -271,6 +271,111 @@
...
+                                "description": "Contains a URI to query creation status."
+                            }
+                        }
+                    },
+                    "default": {
+                        "$ref": "#/components/responses/default"
+                    }
+                }
+            }
+        },
```
cookeac, line 378: Could we also have a /batches/ URL for posting batches of type classifications? I don't think we have this yet unless I've missed it.

erwinspeybroeck: It's not there yet. I propose to tackle this in a seperate action - extra pull request.

cookeac (approved):

AlexeyHardCode (approved):

AlexeyHardCode (approved):

AlexeyHardCode (approved):

AlexeyHardCode (approved):

PR #274, opened 2021-12-16
Fixed configuration for spectral

ahokkonen:
For some reason variable "${GITHUB_REF##*/}" is not resolved properly for ruleset link - added direct urls for ruleset and linting files.

Now linter issue is solved, but there is a still an error during validation on "Unknown value type ty."
Actual error comes from "icarResourceReferenceType.json" in "@type" property. 

Interesting thing - it only fails on docker build of spectral, locally I use the latest one from NPM and it works without any errors...

### DIFF
--- a/.github/workflows/ValidateWithSpectral.yml
+++ b/.github/workflows/ValidateWithSpectral.yml
@@ -2,9 +2,6 @@ name: Spectral JSON Validation CI
 
 on: [push]
 
-env:
-  URL_SCHEMES_PATH: https://raw.githubusercontent.com/${{github.repository}}/${GITHUB_REF##*/}/url-schemes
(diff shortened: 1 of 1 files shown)


cookeac:
This new error looks a lot like https://github.com/stoplightio/spectral/issues/1456 - which has been resolved, and so perhaps a problem with the docker build and versions?

ahokkonen:
That must be it, latest docker package of spectral is from 26th of September. It could be fix is not included.
So this issue could be resolved in the next docker release.

ahokkonen:
We could also use pre-defined GitHub [action](https://github.com/stoplightio/spectral-action) for linter instead of docker image, but current version (0.8.1) is also built using spectral-core 1.4.0.

PR #279, opened 2022-01-14
Feature 266 add resource type

### LINKED ISSUE #266: Add semantic type to resource
The streaming API but many other contexts it is valuable for a resource json fragment to clearly indicate what kind of thing it is.

Proposal:

add a new property to resource called: icarType whose value is a URI. 
define a list of URIs for all existing ICAR types.

The property is optional. 
For the purposes of the protocol this property would need to be present.

This change is backwards compatible.

### LINKED ISSUE #278: Current icarResource Discriminator property doesnt appear valid
The swagger seems incorrect: https://github.com/adewg/ICAR/blob/ADE-1/resources/icarResource.json with relation to the discriminator declaration.

According to the openAPI spec: https://swagger.io/specification/

While composition offers model extensibility, it does not imply a hierarchy between the models. To support polymorphism, the OpenAPI Specification adds the discriminator field. When used, the discriminator will be the name of the property that decides which schema definition validates the structure of the model. As such, the discriminator field MUST be a required field. There are two ways to define the value of a discriminator for an inheriting instance.

Use the schema name.
Override the schema name by overriding the property with a new value. If a new value exists, this takes precedence over the schema name. As such, inline schema definitions, which do not have a given id, cannot be used in polymorphism.

Key thing is that the descriminator property must be on the resource and must be required. From what I can see now this is not the case (unless it is a built in property of object). 

The reason this is brought up is that resourceType is the name of the property we agreed on adding to support the streaming API. The PR to add this property will likely fix this issue. But we should discuss and see if this is really doing something we need and in the way we expect.

gra-moore:
added resourceType to resource.
added catalog and identifiers for all types.


cookeac (changes requested):
This is good. I particularly like the markdown file.
I think the .github/stale.yml change has got caught up in this PR and could be cherry-picked back out?

PR #281, opened 2022-01-26
added location to resource.

### LINKED ISSUE #267: Add location or context to resource
For the streaming API it is necessary to provide context regarding which farm or account the data is related to. 

Options:

1. Add Location as a property of resource. This is backwards compatible. Any resource definitions on subtypes would be moved up to resource. 

2. Add a 'context' property to meta that allows for this to be conveyed in an application independent way.

In either case the API spec would need to declare which properties are required and how they should be interpreted, 

Thoughts: the use of context is a light binding and doesnt change the model at all.

gra-moore:
Added location to the icarResource and removed it from subclasses where it was already defined.

### DIFF
--- a/resources/icarDailyMilkingAveragesResource.json
+++ b/resources/icarDailyMilkingAveragesResource.json
@@ -18,10 +18,6 @@
           "type": "string",
           "description": "Unique identifier in the source system for this event."
         },
-        "location": {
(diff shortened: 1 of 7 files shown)


erwinspeybroeck:
Ok for me

cookeac (approved):

PR #287, opened 2022-04-05
removed discriminators

### LINKED ISSUE #278: Current icarResource Discriminator property doesnt appear valid

gra-moore:


cookeac:
This looks good to me.
- EventType removed from icarEventCoreResource because ResourceType (from icarResource) will serve effectively.
- IdentifierType removed from icarIdentifierType because identifiers were not designed to use inheritance - but it could be replaced if it was really necessary.
- MemberType removed from icarResourceCollection because ResourceType (from icarResource) will serve effectively. Note that having a MemberType is interesting in itself, but as implemented it never actually contained the type of the members, only of the collection.

cookeac (approved):

PR #294, opened 2022-04-21
Attempt to fix Spectral paths

### LINKED ISSUE #270: Fix request for spectral linter
New version of spectral linter requires ruleset parameter.
Creating default ruleset as part of the sources to solve that issue.

cookeac:
The updated spectral code needed a ruleset parameter, which Anton had attempted to specify using an environment variable to choose the appropriate ruleset for Develop/ADE-1, etc. 

This has been broken for a few months, so I am attempting to fix this. According to the documentation, you should use Linux bash syntax, not GitHub syntax when referencing environment variables in the **run** statement. So I am using `$SPECTRAL_RULESET` rather than `{{env.SPECTRAL_RULESET}}`.

### DIFF
--- a/.github/workflows/ValidateWithSpectral.yml
+++ b/.github/workflows/ValidateWithSpectral.yml
@@ -13,11 +13,11 @@ jobs:
     steps:
     - name: validate JSON URL scheme
       run: |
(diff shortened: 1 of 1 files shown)


cookeac:
I'm going to try and merge this, because its not possible to test this without a commit.

Issue #248, opened 2021-09-23
Support linking or referencing related resources


cookeac:
It could be useful to have a generalised method for referencing one event from another, and describing the relationship between the events. This is demonstrated by the desire to link feed intake events and milking events when animals are fed while they are milked #171.

We did something similar for links between spatial resources at https://github.com/Datalinker-Org/Geospatial
There we created a `RelationType` with the properties:

| Name | Type | Description |
| --- | --- | --- |
| relationship | string | How the referenced resource relates to the current resource (using a well-known syntax or FOAF). |
| contentType | string | The name of the resource so a client knows the type of resource that is being referenced. |
| related | string, uri | A URI that may be used to retrieve the related resource. |

We could add to these a `relatedIdentifier` (source and ID?) that could be used in preference to the related URI.

It would then be possible to extend `icarEventCoreResouce` (or even `icarResource`) to have an array of related resources or related events.

alamers:
Sounds interesting. A few questions:
* the 'relationship' type's well-known syntax: this would be some standardised set of types? A quick scan of FOAF indicates that it is defined mostly for a social network context, so this would be up to ADE? I am a bit in doubt in how much a client can expect in this area: do we expect some relations to be compulsory? How do we grow from these 'loosely' defined relations to more strict relations?
* contentType: as in 'text/json'? do we have other content types or could this also refer to things outside ADE?
* related vs relatedIdentifier: I guess this is the semantic web vs 'classic' url construction again? I have a slight preference for id's since then the message is decoupled from its transport but I think we haven't really made a choice in that area.

gra-moore:
I have a few comments:

1. have a well defined section in the entity data structure that contains all references. This allows clients to be sure what group of keys are references.

2. The contentType idea isnt ideal. Instead we should promote the idea of a type property in the data. either via a ref to Class or a promoted key such as rdf:type.

3. Agree the reltype should be a URI and so should the value. If you want to encode schemes then use URNs and map the common agri schemes to URN schemes. 
 
4. The value should be either a single string interpreted as a URI or an array. 

5. This a generalized mechanism for referencing any kinds of things, not just events.

So relationships could look like:

{
   .... 
   "refs" : {
      "http://ade.org/events/parturition" : "urn:myweirdscheme:38835835863"
   }
}  

We should create semantic URIs for the classes and rel types.

cookeac:
Good points:
* Yes, I would see this type being a used in a common array (well-defined section) in `icarResource.json`.
[...]

PR #303, opened 2022-05-19
Create icarGroupEventResource to support event observations on groups

### LINKED ISSUE #273: Add event resources for sets (groups) of animals
A question passed on from members of the ICAR ADE group: should it be possible to record events for sets of animals (groups)? What would be the semantics of this?

I am unsure of the demand for this. Some places where this could be useful:
- Have have applied a standard animal health treatment (for instance a drench or oral wormer) to a group of animals.
- I moved a group of animals to a different sublocation (paddock/field) at a farm location (an event we don't have)

Depending on the semantics, this could just be a shorthand for batch applying an event with the same observed values to many individual animals. Or it might mean that the group of animals was identified prior, and I have applied the event to that group without checking that all the individuals were present -- this would be subtly different.

I would value thoughts or feedback that I could provide as feedback to the ADE group.

### LINKED ISSUE #291: Create icarGroupEventCoreResource
This issue helps to address #273 
* Create icarGroupEventCoreResource.json based on icarEventCoreResource.json once #290 is complete.
* Add properties for group composition and animal set.

Some consideration needs to be given to animal set -- should it be a reference, or should the animal set resource, including its current list of members, be included. A reference is lighter weight, but leaves undefined how to interpret the event if the animal set membership could change.

cookeac:
Defines `icarInventoryClassificationType`, `icarAnimalSetReferenceType`, `icarGroupEventMethodType`.
Extends `icarEventCoreResource` to `icarGroupEventCoreResource`.

Resolves #273 
Resolves #291


types/icarAnimalSetReferenceType.json
```diff
@@ -0,0 +1,16 @@
+{
+    "description": "References an animal set through its unique ID in the source system, optionally also specifying its URL (@ID).",
+
+    "allOf": [
+        { "$ref": "../types/icarResourceReferenceType.json" },
+        {
+            "type": "object",
+            "properties": {
```
AndreasSchultzGEA: no required properties?

types/icarInventoryClassificationType.json
```diff
@@ -0,0 +1,46 @@
+{
+  "type": "object",
+  "description": "This type is used to categorise animals by shared characteristics - so you can say the equivalent of 200 x 2-year-old in-calf Jersey heifers.",
+
+  "properties": {
+    "name": {
+      "type": "string",
+      "description": "Human-readable name for this inventory grouping."
+    },
+    "count": {
```
AndreasSchultzGEA, line 12: does it clash with same property in icarGroupEventCoreResource?

types/icarInventoryClassificationType.json
```diff
@@ -0,0 +1,46 @@
+{
+  "type": "object",
+  "description": "This type is used to categorise animals by shared characteristics - so you can say the equivalent of 200 x 2-year-old in-calf Jersey heifers.",
+
+  "properties": {
```
AndreasSchultzGEA, line 7: nothing required?

resources/icarGroupEventCoreResource.json
```diff
@@ -0,0 +1,35 @@
...
+  "description": "Extends icarEventCoreResource to support event observations on groups of animals.",
+
+  "allOf": [
+    {
+      "$ref": "../resources/icarEventCoreResource.json"
+    },
+    {
+      "type": "object",
+
+      "properties": {
```
AndreasSchultzGEA, line 13: nothing required?

types/icarInventoryClassificationType.json
cookeac: We want to allow users to specify as much or as little as required, but I did wonder if for instance `name` and `species` should be made mandatory. As this is an embedded element, is it likely that some implementers would not be able to leave it out entirely and then be forced to fill in these fields even if using an animal set instead?

resources/icarGroupEventCoreResource.json
cookeac: Its a base class - I think required fields should be specified in concrete classes.

types/icarAnimalSetReferenceType.json
cookeac: Good point. ID should be I think.

types/icarInventoryClassificationType.json
cookeac: My feeling is the count in the event is the number of animals sampled (if not all the animals were observed) whereas here its the number of animals in the group.
I've modified `icarGroupEventCoreResource`, changing `count` to `countObserved` for clarity.

types/icarAnimalSetReferenceType.json
cookeac: But see also #248 which might be a more generalised solution.

resources/icarGroupEventCoreResource.json
erwinspeybroeck: agree

types/icarInventoryClassificationType.json
erwinspeybroeck: Always difficult to really say what is required - mostly it depends on the data providers capabilities and the data consumer needs ...
I would say name would be sufficient?

erwinspeybroeck (approved):

types/icarInventoryClassificationType.json
cookeac: Making name and id required.

types/icarAnimalSetReferenceType.json
cookeac: Changed to align with #248 and made `identifier` required.

resources/icarAnimalEventCoreResource.json
```diff
@@ -1,5 +1,5 @@
 {
-  "description": "",
+  "description": "Extends icarEventCoreResource to add a single individual animal identifier.",
 
   "allOf": [
```
AndreasSchultzGEA, line 4: "animal" as required?

resources/icarGroupEventCoreResource.json
```diff
@@ -0,0 +1,35 @@
+{
+  "description": "Extends icarEventCoreResource to support event observations on groups of animals.",
+
+  "allOf": [
+    {
+      "$ref": "../resources/icarEventCoreResource.json"
+    },
+    {
+      "type": "object",
```
AndreasSchultzGEA, line 9: "groupMethod" as required?
Or isn't at least one groupMethod required?

types/icarAnimalSetReferenceType.json
```diff
@@ -0,0 +1,7 @@
+{
+    "description": "References an animal set through its unique ID in the source system, optionally also specifying its URL (@ID).",
+
+    "required": [ "identifier" ],
```
AndreasSchultzGEA, line 4: identifier is not present

resources/icarAnimalEventCoreResource.json
cookeac: Happy to do this though better to set required fields in concrete implementation classes.

resources/icarGroupEventCoreResource.json
cookeac: Agreed, added.

types/icarAnimalSetReferenceType.json
cookeac: OK. `identifier` was added to icarResourceReferenceType.json. I didn't want to make it required there though, as this would be a breaking change for other users of that type - medicine and device references. Let me know if this causes code generation or implementation problems. Otherwise at the very least it is documentation that we want identifier to be specified.

PR #313, opened 2022-08-25
Address Speccy and Spectral linting errors and warnings

cookeac:
There were a number of Speccy and Spectral errors and warnings that are worth looking into now that Spectral is running again.
In this pull request I've addressed some of these:
- Addressed tag warnings in the URL scheme files, and updated all tags to ADE-1.3-*
- icarHealthStatusObservedEventCollection.json didn't exist - created this
- icarResourceReferenceType.json had "identifier" specified incorrectly using "type" when it should have been "$ref" - corrected.
- Addressed unused section warnings in the URL scheme files - animal-scheme and animal-id only make sense in the registration URL scheme where you can query an animal given its ID. Other schemes don't currently use these parameters (but could?).

There are still errors about `"type" property must be equal to one of the allowed values: "array", "boolean", "integer", "number", "object", "string".` in a number of collection files. This seems to be spurious, and indeed when I reproduce this by running Spectral locally in the VS Code extension, when it hits one of these Spectral stops with an error and directs me to look at its output. The actual JSON Schema looks correct - there is an array, and the error occurs at the start of the "items" clause, not in the "type" clause. It doesn't even work if items is a simple type.


cookeac:
Approved at meeting

PR #315, opened 2022-08-25
Add Group Treatment event

### LINKED ISSUE #314: Add icarGroupTreatmentResource
As referenced in #273, it is advantageous to allow treatments to be recorded on groups of animals, as defined by an animal set or simply by a description (for instance, animals without unique individual identification). 

Examples of these treatments are drenches, dips or pour-ons (wormers) for parasite control, or common vaccinations.

cookeac:
Add a treatment resource with the similar fields to icarTreatmentEventResource, but for a group of animals.
Resolves #314 

Changes from `icarTreatmentEventResource`
- Based on `icarGroupEventCoreResource`
- Removed `Comment` attribute as there is now a `Remark` attribute in icarEventCoreResource that subsitutes for this generally (not removed from icarTreatmentEventResource as that would be a breaking change).
- Replaced `dose` with `dosePerAnimal` to clarify this is the actual or average dose per animal treated in the event.
- Added `totalMedicineUsed` to give the total across all animals treated in the event.

### DIFF
--- a/resources/icarGroupTreatmentEventResource.json
+++ b/resources/icarGroupTreatmentEventResource.json
@@ -1,0 +1,56 @@
+{
+    "description": "Defines a health treatment applied to a group of animals.",
+
+    "allOf": [
+        {
+          "$ref": "../resources/icarGroupEventCoreResource.json"
+        },
+        {
+            "type": "object",
+
+            "properties": {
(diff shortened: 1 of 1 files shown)


cookeac:
Approved at the meeting 8 Sept 22

PR #320, opened 2022-09-09
Add of withdrawal-resources

### LINKED ISSUE #318: Separation of Milk
How to handle the separation of milk in the case of
* parturition -> colostrum
* treatment,
* etc.

AndreasSchultzGEA:


resources/icarWithdrawalResource.json
```diff
@@ -0,0 +1,37 @@
+{
+  "description": "Withdrawal resource to seperate a product.",
```
cookeac:
```suggestion
  "description": "A withdrawal resource used to indicate that product should be separated (e.g. colostrum from newly lactating cows).",
```

resources/icarWithdrawalResource.json
```diff
@@ -0,0 +1,37 @@
...
+  "allOf": [
+    {
+      "$ref": "../resources/icarResource.json"
```
cookeac: Agree this could be a "resource", but it could also be an "event" with the event date being the start of separation, and simply specifying end date and product type (remark being already built into events). 
Either a resource or an event would work -- it just seemed more like another event to me?

cookeac (changes requested):
Suggesting a bit more detail in description to help people understand types of use cases. I also wondered if this should be just a resource or an event - I think either could work.

resources/icarWithdrawalResource.json
AndreasSchultzGEA: changed from Resource to EventResource

cookeac (approved):
Looks good now.

erwinspeybroeck (approved):
also ok for me

cookeac (approved):

cookeac (changes requested):
@AndreasSchultzGEA GitHub says that the branch has conflicts that must be resolved.
The issue is simply that the milkURLScheme.json in this PR has different tab indentation on its lines than that in the Develop branch. Therefore it looks like every line has changed, while in the meantime there have been minor changes to the version in the Develop branch. I think if you reformat this file with the smaller spacing that is in the existing file, it should merge fine.

Also there is a warning on line 519 - the "operation" for "get existing milking-withdrawals" must not have spaces (currently it does have spaces). 

I tried to make the changes and add them to your branch, but of course I don't have permission.

AndreasSchultzGEA:
finally, I re-fixed the formatting

cookeac:
Adds withdrawal resource, resolves #318

PR #321, opened 2022-09-09
changed the default-response to use new icarErrorCollection instead o…

### LINKED ISSUE #295: Usage of openapi-generator 6.0.0-SNAPSHOT
**Setup**: 
- artefacts are not public available, therefore download sources and compile
- compiles only with Java JDK 8. Set Path to JDK and also JAVA_HOME

**generated Java-Sources**:
- introduction of IcarAnimalEventCoreResource causes Problems
- single discriminator located in IcarResource - also defined as dedicated property
`@JsonTypeName("icarResource")
@Generated(value = "org.openapitools.codegen.languages.SpringCodegen", date = "2022-04-25T07:58:00.892882900+02:00[Europe/Berlin]")
public class IcarResource {

  **@JsonProperty("resourceType")
  private String resourceType;**
`

**JSON**:
- ReST-call - e.g. animals, parturitions - contains double definition of 'resourceType'
{
    "view": null,
    "member": [
        {
            "resourceType": "icarReproParturitionEventResource",
            "resourceType": null,
            "@self": null,

### LINKED ISSUE #299: Remove the default response from example URL files and substitute 400 and 500 errors.
Use of the default causes invalid code generation as mentioned in the comments to #295

AndreasSchultzGEA:
…f inline-definition

### DIFF
--- a/collections/icarErrorCollection.json
+++ b/collections/icarErrorCollection.json
@@ -1,0 +1,12 @@
+{
+  "description": "A collection of RFC7807 compliant problem responses for JSON APIs.",
+  "type": "object",
+  "properties": {
+    "errors": {
+      "type": "array",
+      "items": {
+        "$ref": "../resources/icarResponseMessageResource.json"
+      }
+    }
+  }
+}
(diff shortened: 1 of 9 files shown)


cookeac (approved):
This looks good to me.

cookeac:
@AlexeyHardCode Now this is merged can you check it generates ok from C# too?

cookeac:
Resolves #295 (again), Resolves #299

PR #322, opened 2022-09-12
Add PostalAddress to consignment origin and dest

### LINKED ISSUE #317: Consider adding PostalAddress to icarConsignmentType
Currently `icarConsignmentType` has string fields for `destinationAddress` and `originAddress'.
Several systems store addresses in structured fields (e.g. Address 1, Address 2, City, Post Code, Country...). There is a standard for this.

It is relatively easy to convert a structured address to a string -- but much more effort to reconstitute a structured address from a string.

In Schema.Org they get around this by having:

`address: { "anyOf": [ { "string" }, { "$ref": "../types/PostalAddress.json" } `
(this is a simplified approximation)

At face value, this would be a non-breaking change (to validation), but I suspect it would break implementations.

An alternative would be to extend `icarConsignmentType`, adding `destinationPostalAddress` and `originPostalAddress`. Not as elegant, but maybe safer.

cookeac:
Define PostalAddress from schema.org, and add it to icarConsignment's originAddress and destinationAddress. This uses "anyOf" to support backward compatibility for the simple string type. Resolves #317

### DIFF
--- a/types/PostalAddress.json
+++ b/types/PostalAddress.json
@@ -1,0 +1,32 @@
+{
+    "title": "PostalAddress - from schema.org",
+    "description": "An address object from schema.org (see https://schema.org/PostalAddress).",
+    "type": "object",
+  
+    "properties": {
+      "addressCountry": {
+        "description": "The country. For example, USA. You can also provide the two-letter ISO 3166-1 alpha-2 country code.",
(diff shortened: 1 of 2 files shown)


erwinspeybroeck (commented):
looks good - no remarks

Issue #323, opened 2022-09-12
Define URL Scheme for Group Event messages


cookeac:
Should Group Event messages (treatment, arrival, departure, birth, death) for groups of animals that are not individual animal messages be:
* Added to standard URL schemes (e.g. healthURLScheme.json and registrationURLScheme.json) OR
* Added to one or more group URL schemes (e.g. groupEventsURLScheme.json)

I lean towards the second option if there are only a relatively small number of these.

cookeac:
Discussed and agreed:
1. Add these to the health and registration schemes.
2. Add them to exampleURLScheme.json as well

PR #324, opened 2022-09-12
Add Group events for livestock movements

### LINKED ISSUE #316: Add Group Movement/Registration events
These would be particularly useful for countries and software systems that deal with movements of sheep, or indeed pigs.
In most cases these would extend `icarGroupEventCoreResource` and just add the same specific fields that are in the individual animal movement events.

cookeac:
Add group events and collections that support recording:
* Arrivals
* Departures
* Births (registration/start of recording)
* Deaths

Resolves #316

### DIFF
--- a/collections/icarGroupMovementArrivalEventCollection.json
+++ b/collections/icarGroupMovementArrivalEventCollection.json
@@ -1,0 +1,21 @@
+{
+    "description": "Represents a collection of group arrival events. Based on icarResourceCollection to provide paging etc.",
+  
+    "allOf": [{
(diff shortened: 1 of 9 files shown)


erwinspeybroeck (commented):
Andrew, don't we need some example url entries?

No further remarks

cookeac:
> Andrew, don't we need some example url entries?
> 
> No further remarks

I agree - I have addressed this with a question in issue #323 and will do another PR to add these once we have agreed which approach to take.

PR #330, opened 2022-09-22
Erwinspeybroeck

### LINKED ISSUE #326: Check that exampleURLScheme is complete with resource types for 1.3
We think this should be deprecated for the future but in version 1.3 it should contain GET for all resource types.

erwinspeybroeck:
Made the example url consistent with the other urls


url-schemes/exampleUrlScheme.json
```diff
@@ -81,6 +81,45 @@
         }
       }
     },
+    "/locations/{location-scheme}/{location-id}/milking-withdrawals": {
+            "get": {
+                "operationId": "get existing milking-withdrawals",
```
cookeac:
```suggestion
                "operationId": "get-existing-milking-withdrawals",
```

cookeac (changes requested):
One change needed to fix a linting error in exampleUrlScheme.json (we actually need the same change in milkURLScheme.json too, but that is not in this PR).

AlexeyHardCode (approved):

PR #331, opened 2022-10-03
Change icarStatistics value to number

cookeac:
changed to "type": "number", "format": "double", as "double" is not a valid JSON Schema type. Discovered from Speccy validation error in PR #330


erwinspeybroeck (approved):
Fine for me,

Thanks Andrew

PR #333, opened 2022-10-03
Add group weigh event

### LINKED ISSUE #332: Add group event for weighing
Add a group event for weighing (for instance, weighing a sample of animals where IDs are not relevant, or weighing untagged animals).
First proposed in #273

cookeac:
Define a group weigh event that has statistics about a group weighing, and optionally has an array of weights/IDs (we expect mostly this will just be weights or empty, but there's a case for animal IDs if they are returned by scales for a sample of animals in a group).

Added these group events to performanceURLSchema.json

Resolves #332 

Also: Add icarGroupTreatmentCollection.json and add treatment group events to healthURLScheme.json.

### DIFF
--- a/collections/icarGroupTreatmentEventCollection.json
+++ b/collections/icarGroupTreatmentEventCollection.json
@@ -1,0 +1,21 @@
+{
+    "description": "Represents a collection of group health treatment events. Based on icarResourceCollection to provide paging etc.",
+  
+    "allOf": [{
+        "$ref": "../collections/icarResourceCollection.json"
+      },
+      {
+        "type": "object",
+        
+        "properties": {
+          "member": {
+            "type": "array",
+            "items": {
+              "$ref": "../resources/icarGroupTreatmentEventResource.json"
+            },
+            "description": "Provides the array of objects, in this case weighing events."
+          }
+        }
+      }
+    ]
+  }
--- a/resources/icarGroupWeightEventResource.json
+++ b/resources/icarGroupWeightEventResource.json
@@ -1,0 +1,47 @@
+{
+    "description": "The Group Weight event records liveweight observations for a group of animals",
+    "allOf": [
+        {
+            "$ref": "../resources/icarGroupEventCoreResource.json"
+        },
+        {
+            "type": "object",
+            "properties": {
+                "units": {
+                    "$ref": "../enums/uncefactMassUnitsType.json",
+                    "description": "Units specified in UN/CEFACT 3-letter form. Default if not specified is KGM."
+                },
+                "method": {
+                    "$ref": "../enums/icarWeightMethodType.json",
+                    "description": "The method of observation. Loadcell is the default if not specified."
+                },
+                "resolution": {
+                    "type": "number",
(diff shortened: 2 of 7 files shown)


erwinspeybroeck (approved):
looks logical to me - more of the same stuff.
Didn't dive into it very deep

collections/icarGroupTreatmentEventCollection.json
```diff
@@ -0,0 +1,21 @@
...
+      {
+        "type": "object",
+        
+        "properties": {
+          "member": {
+            "type": "array",
+            "items": {
+              "$ref": "../resources/icarGroupTreatmentEventResource.json"
+            },
+            "description": "Provides the array of objects, in this case weighing events."
```
AndreasSchultzGEA, line 16: treatment instead of weighing

resources/icarGroupWeightEventResource.json
```diff
@@ -0,0 +1,47 @@
+{
+    "description": "The Group Weight event records liveweight observations for a group of animals",
+    "allOf": [
+        {
+            "$ref": "../resources/icarGroupEventCoreResource.json"
+        },
+        {
+            "type": "object",
+            "properties": {
+                "units": {
```
AndreasSchultzGEA, line 10: singular?

resources/icarGroupWeightEventResource.json
```diff
@@ -0,0 +1,47 @@
...
+                    "description": "The smallest measurement difference that can be discriminated given the current device settings. Specified in Units, for instance 0.5 (kilograms)."
+                },
+                "animalWeights": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "../types/icarIndividualWeightType.json"
+                    },
+                    "description": "Array of animal id and weight pairs for animals in the event."
+                },
+                "statistics": {
```
AndreasSchultzGEA, line 29: I suppose this to relate to animalWeights. 
But unsure why this is an Array!?

collections/icarGroupTreatmentEventCollection.json
cookeac: Correct - an afterthought but see the bottom line of the PR description. I went to update the collections and noted the the group treatment event collection didn't exist.

resources/icarGroupWeightEventResource.json
cookeac: a) English would only use "unit" if the value was singular, which it mostly will not be. Also like this for consistency with `icarMassMeasureType.units`.

resources/icarGroupWeightEventResource.json
cookeac: Two arrays in this resource:
* An array of statistics (e.g. min, max, mean, standard deviation, coefficient of variation, or anything else you like
* An array of raw data, which coming from devices may have just weights, or some values with IDs (which I'd prefer were individual animal events, but sometimes are not). The raw data (if provided) is what was analysed to produce the statistics for the event.

cookeac:
@donwilson-rezare FYI

PR #334, opened 2022-10-04
Inventory classification change - fix birthPeriod validation

### LINKED ISSUE #335: Birth Period in icarInventoryClassificationType doesn't validate as per description
The `birthPeriod` in `icarInventoryClassificationType` was intended to represent an approximate birth period for a number of animals with an **ISO8601 interval** (effectively a range between two dates, but there are extensions that allow year or year and month).
I had originally drafted this as an `icarDateType` which is just **string format date**. However, most validators only accept RFC-3339 dates in this context, and not intervals.

I propose to make this a string, with a pattern that reflects the description.

cookeac:
birthPeriod was previously specified as an icarDateType, which is half-right, because it is a date **range**.
Most validators interpret "format": "date" as an RFC3339 date, which does not allow for ranges.

Instead I have changed its type to string, and defined a pattern with a regex. If this is too complex we can ditch the regex and simply make it a string with the current description.

### DIFF
--- a/types/icarInventoryClassificationType.json
+++ b/types/icarInventoryClassificationType.json
@@ -26,8 +26,10 @@
       "description": "Primary breed defined using an identifier and scheme."
     },
     "birthPeriod": {
-      "$ref": "../types/icarDateType.json",
-      "description": "The range of birth dates. Use YYYY (all one year), YYYY-MM (one month), or two RFC3339 date-times separated by / to represent a range."
+      "type": "string",
+      "pattern": "^(([0-9]{4})|([0-9]{4}-[0-9]{2})|([0-9]{4}-[0-9]{2}-[0-9]{1,2}(Z?)(\/|--)[0-9]{4}-[0-9]{2}-[0-9]{1,2}(Z?)))$",
+      "description": "The range of birth dates. Use YYYY (all one year), YYYY-MM (one month), or two RFC3339 date-times separated by / to represent a range.",
+      "nullable": true
     },      
     "reproductiveStatus": {
       "$ref": "../enums/icarAnimalReproductionStatusType.json",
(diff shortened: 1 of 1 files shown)


erwinspeybroeck (approved):
No comment from my side

AndreasSchultzGEA (changes requested):
* the 2 parts differ. whereas the first one is very flexible, the second one is fixed to YYYY-MM-DDZ
* the first part offers the possibility to insert YYYY-99, e.g. by defining [0-9]{2} as month
* Is the second part really optional? I think not. The separator as well.

cookeac:
> 

Having tested this in https://regex101.com/ there are three effective captures:
* `([0-9]{4})` - four digits (anything from 0000 to 9999 but effectively designed as year)
* `([0-9]{4}-[0-9]{2})` - four digits, a dash, and 2 digits. Intended to represent YYYY-MM but as you say it does allow for 9999-99. I will make a change to allow 00 to 19 for months (because more than that makes the regex complex).
* `([0-9]{4}-[0-9]{2}-[0-9]{1,2}(Z?)(\/|--)[0-9]{4}-[0-9]{2}-[0-9]{1,2}(Z?)))` - intended to be YYYY-MM-DDZ/YYYY-MM-DDZ
The Zs are optional (indicated by the question mark after them), but we could remove them completely. The only optionality then is whether you have a / or -- symbol. The latter is an alternative more suited to some operating systems.

image

So I think the optionality is right, but would be willing to accept corrections :-)
The remaining questions are:
* Do we even want the Z given that we just have dates and not times? I will remove if not.
* Do we want more sophisticated regex that enforces the exact day and month syntax? Otherwise I will just make changes that get us in the right approximate range - for instance 00-19 month and 00-39 days. This is easy.

AndreasSchultzGEA (approved):

Issue #300, opened 2022-05-19
Document the ADE information model


cookeac:
Create model documentation including diagrams

cookeac:
For instance, replace https://github.com/adewg/ICAR/wiki/Working-with-Resources

cookeac:
Relates to #258 but probably initially more informal and diagrammatic.

gra-moore:
@cookeac We have this issue and another related which is to create a formal model. I think these things are almost the same. Defining the underlying data model would benefit understanding of whats in place, facilitate alternative semantic syntax representations. This would benefit the current release and provide a way to make this version work in the semantic tech world. Suggest we start work on this.

cookeac:
Components:
* Landing page(s) and maybe consider GitBook etc.
* Ontology or semantic data dictionary --> Graham is willing to make a start on this so others can contribute
* Information model document (with some diagrams)
* Automated generation from the JSON schema?
* Getting started guide (implementation)
* Getting started guide - overview of the specification

Agreed to
1. Create issues for each of the above
2. Find someone to lead each
3. Ask our organisations for volunteers to contribute or proofread

PR #338, opened 2022-10-20
Documentation for group events

### LINKED ISSUE #272: Document semantics for Animal Sets - Groups
Feedback from the ICAR ADE group is that it is not clear whether the need to manage groups of animals could be met with the Animal Sets functionality we have defined. I believe that could be (or at least mostly - we would need to understand if there were specific problem cases).

However, it sounds like we need to create some documentation so that the use of Animal Sets for groups is discoverable.

cookeac:
Preview of documentation (much more to be done)
- Explanation of animal sets and groups (resolves #272)
- Start of ADE information model documentation (#300)


cookeac:
Simplified - just a few readme files and then the documentation of when to use animal sets and when to use group events.
The internal links in the latter document now point directly to the resource JSON schemas (rather than to manually maintained markdown or trying to deeplink into a open API schema viewer.

tpekeler (approved):
New structure seems to be much better maintainable. Differences between Animal sets and Group events seems to be described clearly. Cannot verify the links embedded, since they are not opening correctly, needs to be confirmed, that links are correct after merging.

Issue #285, opened 2022-03-09
Provide predictions for expected milk yield


dirbSEGES:
Hi, 

Could TestDayResultEventResource have a new property:   "milkWeightExpected" or "milkWeightGoal" ? 

BR

cookeac:
We agree we should add `milkWeight24HoursPredicted`.
It would also be useful to consider how predictions are interchanged more generally.

cookeac:
Erwin will investigate this.

erwinspeybroeck:
1) it is possible to add predicted fields to the test day result API - on our database they are called:
- milkWeightExpected
- fatWeightExpected
- ProteinWeightExpected
Andrews proposal is maybe better since it makes clear that it 's a 24 hours expected weight
- milkWeight24HoursPredicted
- fatWeight24HoursPredicted
- proteinWeight24HoursPredicted

2) we deliver now to feeding companies also milk-predictions
This contains in fact information (milk, fat and protein) about the expected lactation curve:
- expected average 24  hour production for the rest of the lactation (end based on expected calving date)
- if the peak of the lactation still has to occur : the exact days in milk the peak will be reached and the predicted 24 h production on that peak
- the predicted 24h production on the next (in the future) milk recording (if no info about when - 3 weeks is taken as interval

Do we need to define an ICAR standard API for this?  I will need it to phase out some old software. I might define it for now in the way we want to provide this information to our customers. Has naybody else such kind of information they share at this moment?

cookeac:
We discussed at our meeting on 2022-08-25 and agreed that it would make sense to have a message such as that described in option 2 - a more generalised lactation prediction for individual animals. We wondered if this might avoid having to have prediction fields in the test day resource. However, if the lactation prediction was always future-focused, you could then lose the previous prediction for the test day, and not be able to compare it.
@erwinspeybroeck will start to model this up (code, diagram, or documentation - any is fine) and we will review and discuss at a future meeting.

erwinspeybroeck:
I did the following:

- added predicted fields to the test day result resource
- added a new resource "milkPrediction"
- added it for the moment in the milk url only

It can be viewed by https://editor.swagger.io/?url=https://raw.githubusercontent.com/adewg/ICAR/erwinspeybroeck/url-schemes/milkURLScheme.json

cookeac:
On 2022-11-03 @erwinspeybroeck took us through the excellent and very interesting API design for this.
We agreed:
[...]

PR #377, opened 2023-03-23
#285 milk predictions

erwinspeybroeck:
[https://github.com/adewg/ICAR/issues/285](url)

Adds milk prediction fields in the test day results resource

Adds a new API for milk predictions


resources/icarMilkPredictionResource.json
```diff
@@ -0,0 +1,27 @@
+{
+  "description": "This returns a prediction of the expected milk, fat and protein for a cow for the remaining the lactation. Also a prediction of this production on the next testday is included and it eventually gives indication of when the peak production is reached.",
```
cookeac:
```suggestion
  "description": "This returns a prediction of the expected milk, fat and protein for a cow for the remaining the lactation. A prediction of this production on the next testday is included and it also gives indication of when the peak production is reached. The eventDateTime for this event is the date of the prediction is made.",
```

cookeac: I've suggested clarifying in the description how the event date works (to avoid confusion). Hopefully this is correct!

types/icarMilkingPredictionType.json
```diff
@@ -0,0 +1,26 @@
...
+  "required": [
+    "unitCode",
+    "value"
+  ],
+
+  "properties": {
+    "milkWeight": {
+      "$ref": "../types/icarMilkingMilkWeightType.json"
+    },
+    "fatWeight": {
```
cookeac, line 15: Do we actually need icarMilkingFatWeightType and icarMilkingProteinWeightType types - or could we just use the existing icarMilkingMilkWeightType which has all the same fields and enumerations? The fields are already differentiated by their names (fatWeight, proteinWeight) - I don't think they need to be different types as well.

types/icarMilkingProteinWeightType.json
cookeac, line 1: This file may not be needed - could just use icarMilkingMilkWeightType in its place.

types/icarMilkingFatWeightType.json
cookeac, line 1: This file may not be needed - could just use icarMilkingMilkWeightType in its place.

url-schemes/milkURLScheme.json
```diff
@@ -376,6 +376,45 @@
                 }
             }
         },
+        "/locations/{location-scheme}/{location-id}/milk-predictions": {
+            "get": {
+                "operationId": "get-milk-predictions",
+                "summary": "Get the milk predictions for a certain location",
+                "description": "# Purpose\nProvides the milk predictions for a location\n",
+                "tags": [
+                    "ADE-1.4-milk"
```
cookeac, line 385:
```suggestion
                    "ADE-1.3-milk"
```

cookeac: Keep tags consistent for now to avoid linting errors. We will bulk update tags before releasing 1.4.

cookeac (changes requested):

PR #381, opened 2023-04-03
Implement Milking Predictions

erwinspeybroeck:

PR #385, opened 2023-04-12
Inventory transaction resources

### LINKED ISSUE #369: Add Inventory to MedicineRessource?
Hi 

I will suggest to add "inventory" to the MedicineRessource. To handle the inventory of medicine in the given location.  (I haven´t been able to find a ressource to handel this?)

add something like: 

"inventory": {
                    "type": "string", --?? maybe number, it can be a negative inventory.
                    "description": "Inventory of the medicine at this actuelly time"
 
and also the unit have to be added: 

"DoseUnits": {
            "$ref": "../enums/uncefactDoseUnitsType.json",
            "description": "Units of measurement in UN/CEFACT 3-letter form"

cookeac:
Defines a generalised icarInventoryTransactionResource and its specific versions, icarFeedTransactionResource and icarMedicineTransactionResource. Resolves #369

Note - update to exampleURLScheme included, but I know that I will need to merge these changes with the changes in #381 before this can be merged.

### DIFF
--- a/types/icarFeedReferenceType.json
+++ b/types/icarFeedReferenceType.json
@@ -1,20 +1,22 @@
 {
     "description": "Feed Reference defines a feed product.",
-    "type": "object",
-    "$id": "types/icarFeedReferenceType",
+
     "allOf": [
         {
             "$ref": "../types/icarProductReferenceType.json"
-        }
-    ],
-    "properties": {
-        "category": {
-            "$ref": "https://raw.githubusercontent.com/Pure-Farming/ICAR/ADE-1/enums/icarFeedCategoryType.json",
-            "description": "Defines the category of the feed product."
         },
-        "type": {
-            "$ref": "https://raw.githubusercontent.com/Pure-Farming/ICAR/ADE-1/types/icarFeedIdentifierType.json",
-            "description": "The scheme + id identifying the type of feed."
+        {
+            "type": "object",
+            "properties": {
+                "category": {
+                    "$ref": "https://raw.githubusercontent.com/Pure-Farming/ICAR/ADE-1/enums/icarFeedCategoryType.json",
+                    "description": "Defines the category of the feed product."
+                },
+                "type": {
+                    "$ref": "https://raw.githubusercontent.com/Pure-Farming/ICAR/ADE-1/types/icarFeedIdentifierType.json",
+                    "description": "The scheme + id identifying the type of feed."
+                }
+            }
         }
-    }
+    ]
 }
--- a/types/icarFeedReferenceType.json
+++ b/types/icarFeedReferenceType.json
@@ -9,7 +9,7 @@
             "type": "object",
             "properties": {
                 "category": {
-                    "$ref": "https://raw.githubusercontent.com/Pure-Farming/ICAR/ADE-1/enums/icarFeedCategoryType.json",
+                    "$ref": "../enums/icarFeedCategoryType.json",
                     "description": "Defines the category of the feed product."
                 },
                 "type": {
--- a/types/icarFeedReferenceType.json
+++ b/types/icarFeedReferenceType.json
@@ -13,7 +13,7 @@
                     "description": "Defines the category of the feed product."
                 },
                 "type": {
-                    "$ref": "https://raw.githubusercontent.com/Pure-Farming/ICAR/ADE-1/types/icarFeedIdentifierType.json",
+                    "$ref": "../types/icarFeedIdentifierType.json",
                     "description": "The scheme + id identifying the type of feed."
                 }
             }
--- a/types/icarFeedReferenceType.json
+++ b/types/icarFeedReferenceType.json
@@ -1,0 +1,20 @@
+{
+    "description": "Feed Reference defines a feed product.",
+    "type": "object",
+    "$id": "types/icarFeedReferenceType",
+    "allOf": [
+        {
+            "$ref": "../types/icarProductReferenceType.json"
+        }
+    ],
+    "properties": {
+        "category": {
+            "$ref": "https://raw.githubusercontent.com/Pure-Farming/ICAR/ADE-1/enums/icarFeedCategoryType.json",
+            "description": "Defines the category of the feed product."
+        },
+        "type": {
+            "$ref": "https://raw.githubusercontent.com/Pure-Farming/ICAR/ADE-1/types/icarFeedIdentifierType.json",
+            "description": "The scheme + id identifying the type of feed."
+        }
+    }
+}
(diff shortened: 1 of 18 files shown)


types/icarInventoryTransactionType.json
```diff
@@ -0,0 +1,58 @@
...
+            "properties": {
+                "transactionKind": {
+                    "$ref": "../enums/icarInventoryTransactionKindType.json",
+                    "description": "Identifies the transaction kind."
+                },
+                "quantity": {
+                    "type": "number",
+                    "description": "The overall volume, weight or count of the product in the transaction in the units defined."
+                },
+                "units": {
```
AndreasSchultzGEA, line 27: Singular instead of plural?

types/icarProductReferenceType.json
```diff
@@ -0,0 +1,44 @@
...
+        {
+            "type": "object",
+
+            "required": [
+                "identifier",
+                "family"
+            ],
+
+            "properties": {
+                "identifiers": {
```
AndreasSchultzGEA, line 18: from icarResourceReferenceType, there is the included identifier-property of type icarIdentifierType. 
Shouldn't the array better be named alternativeIdentifier as everywhere else? 
And: The type of identifier and the array of alternativeIds differ (icarProductIdentifierType).
Ok, the content is the same. But still different types.

types/icarMedicineReferenceType.json
```diff
@@ -2,15 +2,12 @@
     "description": "Provides basic details about a medicine and links to a medicine resource (if available).",
     "allOf": [
         {
-            "$ref": "icarResourceReferenceType.json"
+            "$ref": "../types/icarProductReferenceType.json"
```
AndreasSchultzGEA, line 5: may be a breaking change. But only new properties are added by the changes inheritance. Unsure!

erwinspeybroeck: Andrew, a general remark from my side.
You have a generic API for inventory-transactions: GET /locations/{location-scheme}/{location-id}/inventory-transactions and specific ones for feed and medicines. But in the specific ones also the complete list of family products is taken _[ Animal Feeds, Animal Reproductive Products, Veterinary Supplies, Seed and Plant Material, Fertilisers and Nutrients, Pest Control Products, Other Animal Products, Milking Supplies, Fencing Supplies, Water System Supplies ]_, while one would expect this to be limited to e.g. Animal Feeds in case of feed. 
Not a blocking remark, I don't know if we can have a fixed family ...

erwinspeybroeck:
Andrew, a general remark from my side.
You have a generic API for inventory-transactions: GET /locations/{location-scheme}/{location-id}/inventory-transactions and specific ones for feed and medicines. But in the specific ones also the complete list of family products is taken _[ Animal Feeds, Animal Reproductive Products, Veterinary Supplies, Seed and Plant Material, Fertilisers and Nutrients, Pest Control Products, Other Animal Products, Milking Supplies, Fencing Supplies, Water System Supplies ]_, while one would expect this to be limited to e.g. Animal Feeds in case of feed. 
Not a blocking remark, I don't know if we can have a fixed family ...

cookeac:
> Andrew, a general remark from my side. You have a generic API for inventory-transactions: GET /locations/{location-scheme}/{location-id}/inventory-transactions and specific ones for feed and medicines. But in the specific ones also the complete list of family products is taken _[ Animal Feeds, Animal Reproductive Products, Veterinary Supplies, Seed and Plant Material, Fertilisers and Nutrients, Pest Control Products, Other Animal Products, Milking Supplies, Fencing Supplies, Water System Supplies ]_, while one would expect this to be limited to e.g. Animal Feeds in case of feed. Not a blocking remark, I don't know if we can have a fixed family ...

You're exactly right, and I was aware of this. The only thing I could think was documentation - that if you were using the feed or medicine resource type, the family should be set to a constant value. I'll ask my team if there is any way to use the "default" keyword if the property is already defined.

types/icarInventoryTransactionType.json
cookeac: In English we would almost always use plural if the value/amount measure could be greater than one. An alternative is unitCode, which is more descriptive but not consistent with other places (e.g. icarMassMeasureType.json).

types/icarMedicineReferenceType.json
cookeac: As you say, only new properties are added, and when parsed by code generation the result should be consistent. I also ensured that the new properties are not required.

types/icarProductReferenceType.json
cookeac: Noted that we do use `alternativeIdentifiers` in locations and animals. In those cases there tends to be one formal identifier and several alternatives. I'm less convinced with products, medicines, etc where there seem to be a range of identifiers. The source system has obviously chosen one to be its `identifier`, but presumably the destination may use another. Happy to change this to `alternativeIdentifiers` for consistency, but not sure it is needed.

Noted on the types - but that is a feature of the base `icarResourceReferenceType` class and was already the case with medicines, for instance.

AndreasSchultzGEA (commented):
Hello Andrew, I'm sorry but I couldn't finish the review due to missing time.

types/icarFeedReferenceType.json
erwinspeybroeck, line 1: a bit strange to see the url with Pure-Farming in here

erwinspeybroeck (approved):

types/icarFeedReferenceType.json
cookeac: Ah, I didn't catch making that one relative. Will fix.

types/icarFeedReferenceType.json
```diff
@@ -0,0 +1,22 @@
...
+
+    "allOf": [
+        {
+            "$ref": "../types/icarProductReferenceType.json"
+        },
+        {
+            "type": "object",
+            "properties": {
+                "category": {
+                    "$ref": "https://raw.githubusercontent.com/Pure-Farming/ICAR/ADE-1/enums/icarFeedCategoryType.json",
```
cookeac:
```suggestion
                    "$ref": "../enums/icarFeedCategoryType.json",
```

types/icarFeedReferenceType.json
```diff
@@ -0,0 +1,22 @@
...
+                    "description": "Defines the category of the feed product."
+                },
+                "type": {
+                    "$ref": "https://raw.githubusercontent.com/Pure-Farming/ICAR/ADE-1/types/icarFeedIdentifierType.json",
```
cookeac:
```suggestion
                    "$ref": "../types/icarFeedIdentifierType.json",
```

cookeac:
Reviewed at previous meetings; merging.

PR #398, opened 2023-06-28
Changes to the icarReproHeatEvent to make it device independent

### LINKED ISSUE #354: Heat recording - standardise vendor-specific fields and make them extensible
We noted that icarReproHeatEventResource has vendor-specific fields for:
- heatReportNedapCowControl
- heatReportScrSenseTime
These have similar data and the full set of data could encompass both fields.

We wondered if there are ICAR guidelines for standard values.
We would like to consider making standard and/or extensible fields in future and deprecating these ones.

cvigorsICBF:


cookeac (changes requested):
@cvigorsICBF we agreed at the meeting on 29 Jun 2023 that this looks great.
One suggestion please - add `"deprecated": true` to both the `heatReportScrSenseTime` and `heatReportNedapCowControl` sections, so that people know these should no longer be used.

cookeac (approved):

erwinspeybroeck:
No extra remarks from my side.

erwinspeybroeck (approved):

PR #423, opened 2023-09-26
fix: value added to icarInventoryTransactionKindType

### LINKED ISSUE #422: Extend feed inventory transactions to allow feeds to be manufactured or harvested on farm
The existing `icarFeedTransactionResource` represents a transaction that can be used to track receipt or despatch/disposal of feed inventory items at a location (farm) level. The types of transaction are enumerated in `icarInventoryTransactionKindType` and work for a variety of inventory items including medicine and reproductive materials.
```
[  "Receipt", "Disposal", "OnHand", "StockTake", "Use" ]
```

However, it is possible that livestock feed may be conserved and harvested on farm (for instance, hay, straw, and silage). To support this being visualised through the same feed transaction resource, we could add `"Produce"` to the enumeration as a term for on-farm manufacture or harvest of a feed.

@ErikUniformAgri, @erwinspeybroeck you might have some thoughts on this?
@donwilson-rezare tagged for reference.

donwilson-rezare:
Added value "Produce" to the enumeration icarInventoryTransactionKindType

### DIFF
--- a/enums/icarInventoryTransactionKindType.json
+++ b/enums/icarInventoryTransactionKindType.json
@@ -5,6 +5,7 @@
         "Receipt",
         "Disposal",
         "OnHand",
+        "Produce",
         "StockTake",
         "Use"
     ]


cookeac:
Resolves #422

cookeac (approved):
Approving subject to feedback on issue #422

PR #427, opened 2023-10-25
Add resourceType column to resourceTypeCatalog.md

### LINKED ISSUE #426: Add Discriminator to resourceTypeCatalog.md
The file [resourceTypeCatalog.md](https://github.com/adewg/ICAR/blob/Develop/resources/resourceTypeCatalog.md) defines a short name and URI for use in describing resources. However, it could also cause confusion because the OpenAPI discriminator for a resource (`resourceType`) is the same name as the JSON file for each resource without the file extension. This is described in [the generic data exchange API specification](https://github.com/adewg/ICAR/blob/Develop/docs/generic-data-exchange-api.md), but it is easy to miss if not looking at this file.

Proposal: Add a "Discriminator (resourceType)" column to the resource type catalog markdown file.

cookeac:
Add a resourceType column to the ICAR Resource Type Catalog list, to avoid confusion about what value should be implemented in the OpenAPI `resourceType` discriminator. Also add the new resource types that are in the Develop branch.

Resolves #426

### DIFF
--- a/resources/resourceTypeCatalog.md
+++ b/resources/resourceTypeCatalog.md
@@ -2,7 +2,7 @@
 
 This document defines a set of URI identifiers and string short codes and maps them to ICAR resource types. 
 
-The catalog is represented as a table with the columns, shortname, URI, link to resource resource type in github.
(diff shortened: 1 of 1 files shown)


erwinspeybroeck (approved):
Approved!

PR #430, opened 2023-11-16
Define position observations

### LINKED ISSUE #383: Animal and Group Position events
I have had requests for an event that identifies an animal or group of animals observed at a location. 
This might be:
- An animal is identified by an EID reader in a very extensive farming operation where animals are only seen a few times a year. This confirms that the animal is still alive and was seen in that location.
- An ear tag or collar device locates an animal at a certain position so we know it was in a particular barn stall or outdoor paddock
- A farmer records that a group of animals (animal set or inventory classification) was grazing in a particular paddock.

cookeac:
Define position observation events (group and animal) that allow you to record that an animal (or group/mob) was seen (observed) at a given named position or geospatial location.

Resolves #383

Note: #383 talked about these as "location" events originally, but this could cause confusion with Location = Farm/Holding.
In this context we mean a barn, pen, field, paddock, or other named area, or a geospatial location (a polygon or a point such as a lat/long position).

Note: This PR references the GeoJSON `geometry` object defined at https://geojson.org/schema/Geometry.json
This currently causes an OAS3 parsing error, because the GeoJSON schema uses the $schema and #id keywords that are not supported in OpenAPI 3.0. They will be supported when we upgrade to OpenAPI 3.1 and the error should disappear.

### DIFF
--- a/types/icarPositionObservationType.json
+++ b/types/icarPositionObservationType.json
@@ -8,6 +8,10 @@
       "type": "string",
       "description": "The name of a location, such as a barn, pen, building, or field."
     },
+    "site": {
+      "type": "string",
+      "description": "Identifier for a sorting site (icarSortingSiteResource) for this position."
+    },
     "geometry": {
       "$ref": "https://geojson.org/schema/Geometry.json",
       "description": "A GeoJSON geometry (such as a latitude/longitude point) that specifies the position."
--- a/types/icarPositionObservationType.json
+++ b/types/icarPositionObservationType.json
@@ -1,0 +1,16 @@
+{
+  "description": "This type may be included in a position observation event to identify either a named position (such as a barn or pen) or a geographic location.",
+
+  "type": "object",
+
+  "properties": {
+    "positionName": {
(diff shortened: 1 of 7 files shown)


cookeac:
Updated position observations to support the site ID from sorting sites. See also #456.

AndreasSchultzGEA (approved):

types/icarPositionObservationType.json
```diff
@@ -0,0 +1,20 @@
...
+  "properties": {
+    "positionName": {
+      "type": "string",
+      "description": "The name of a location, such as a barn, pen, building, or field."
+    },
+    "site": {
+      "type": "string",
+      "description": "Identifier for a sorting site (icarSortingSiteResource) for this position."
+    },
+    "geometry": {
```
AlexeyHardCode, line 15: I was a little bit worried how this property will be generated in the c# code, but seems that there is nothing special about it. Works fine. 
image

AlexeyHardCode (approved):
Changes looks fine. Code generation works as expected.

PR #431, opened 2023-11-17
Additional data for IcarMovementArrivalEventResource

### LINKED ISSUE #432: IcarMovementArrivalEventResource: additional information about former events
In our HM-System and at least in one of our partners, there is the request to enter additional information about an arriving animal which has taken place bevor the ArrivalEventDateTime.
[PR](https://github.com/adewg/ICAR/pull/431)

AndreasSchultzGEA:
In our HM-System and at least in one of our partners, there is the request to enter additional information about an arriving animal which has taken place bevor the ArrivalEventDateTime.

### DIFF
--- a/resources/icarAnimalCoreResource.json
+++ b/resources/icarAnimalCoreResource.json
@@ -81,22 +81,6 @@
           "$ref": "../enums/icarAnimalLactationStatusType.json",
           "description": "Lactation status of the animal."
         },
-        "parity": {
-          "type": "number",
-          "description": "The parity of the animal during this lactation."
-        },
-        "lastCalvingDate": {
-          "$ref": "../types/icarDateType.json",
-          "description": "RFC3339 UTC date (see https://ijmacd.github.io/rfc3339-iso8601/)."
-        },
-        "lastInseminationDate": {
-          "$ref": "../types/icarDateType.json",
-          "description": "RFC3339 UTC date (see https://ijmacd.github.io/rfc3339-iso8601/)."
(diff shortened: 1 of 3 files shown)


erwinspeybroeck (commented):
Andreas, I'm a bit reluctant to do so. There are other API's to get that information, and what about last lactation, last lactation index, last treatment, ....
If the fields are not mandatory and nobody expects them from us, then it is of course possible ...

cookeac:
I acknowledge some of this is covered by a lactation, but that this may not be immediately available to the destination farm during a movement. I'm not convinced that the solution is to add these fields to the movement event. We should look at adding them as optional status extensions to `icarAnimalCoreResource`, which is supplied as the `animalDetail` field of the movement.
I'll put this as a comment in #432 as well.

resources/icarAnimalCoreResource.json
```diff
@@ -81,6 +81,22 @@
           "$ref": "../enums/icarAnimalLactationStatusType.json",
           "description": "Lactation status of the animal."
         },
+        "parity": {
```
cookeac:
```suggestion
        "currentLactationParity": {
```

AndreasSchultzGEA: @cookeac, I changed as agreed on.

cookeac (approved):

erwinspeybroeck (approved):

cookeac:
Resolves #432

PR #433, opened 2023-11-27
added 2nd managementTag

AndreasSchultzGEA:
There is sometimes the need for space of a second management-tag, e.g. in the case of a neck- and foot-responder

### DIFF
--- a/resources/icarAnimalCoreResource.json
+++ b/resources/icarAnimalCoreResource.json
@@ -57,6 +57,10 @@
           "type": "string",
           "description": "The identifier used by the farmer in day to day operations. In many cases this could be the animal number."
         },
+        "managementTag2": {
+          "type": "string",
+          "description": "The 2nd identifier used by the farmer in day to day operations. In many cases this could be the animal number."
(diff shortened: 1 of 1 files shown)


erwinspeybroeck (commented):
I kow we have another understanding about the management  tag. For me it was the small animal number that the farmer uses in his day to day operations. The respondernumbers could also be added as an alternative identifier, and that is an array.
Managementtag could also become an array, but that would be a breaking change.
Wondering what others are using and how they look at this.

cookeac:
> I kow we have another understanding about the management tag. For me it was the small animal number that the farmer uses in his day to day operations. The respondernumbers could also be added as an alternative identifier, and that is an array. Managementtag could also become an array, but that would be a breaking change. Wondering what others are using and how they look at this.

I agree - additional device identifiers for animals (radio frequency devices tags, ankle or neck or bolus responders) should use the existing `alternativeIdentifiers[]` array with an appropriate scheme. The `managementTag` was defined for a short farmer ID.

PR #438, opened 2023-12-01
Update to OpenAPI version 3.1

### LINKED ISSUE #239: Moving to open api spec 3.1
Related to #214, the question pops up whether or not to move our spec to OpenAPi 3.1. 

Two main questions on this topic:
* Does this impact (runtime) backwards compatibility? (expectation: no)
* What is our policy on following the OpenApi spec? What can our partners expect with regards to maintaining a stable tool stack?

### LINKED ISSUE #451: Remove "enum" attribute from deprecated "@type" property of icarResourceType
The "enum" keyword in this context breaks code generation, and its in a deprecated property, so we will remove it.

cookeac:
Update to OpenAPI 3.1

The following changes have been made to all OpenAPI files in the url-schemes folder:

* Updated to `"openapi": "3.1.0"`
* Updated to version 1.4 (ahead of the ADE 1.4 release)
* Updated all tags to ADE-1.4-xxxx

The following changes have been made to JSON Schema files in the types and resources folders where "nullable": true was previously specified:

* replace `"type": "string"` with `"type": ["string", "null"]`
* replace `"type": "number"` with `"type": ["number", "null"]`
* replace `"type": "integer"` with `"type": ["integer", "null"]`
* replace `"type": "boolean"` with `"type": ["boolean", "null"]`
* replace `"type": "array"` with `"type": ["array", "null"]`

Where a complex type (imported with "$ref" was previously nullable):
* If the type was an object, add `"type": ["object", "null"]` before the $ref
* If the type was a string (an enumeration or an icarDateTimeType), add `"type": ["string", "null"]` before the $ref

These last changes appear in examples online, but they are the ones that are most likely to interrupt code generation. Please test.

Resolves https://github.com/adewg/ICAR/issues/239


cookeac:
A further note - until it is merged, this PR will fail Speccy validation. This is because Speccy is deprecated and does not support OpenAPI 3.1. When this PR is merged, it will remove the GitHub action for Speccy, leaving us with Spectral validation.

cookeac:
We agreed (meeting 2024-04-04) that we could merge this once #451 is addressed with a PR, as @AlexeyHardCode has tested code generation with C# at least.

PR #442, opened 2024-01-25
Make icarDeviceResource an icarResource

### LINKED ISSUE #439: icarDeviceResource is not correctly structured
This resource does not inherit from icarResource - but it should be.

cookeac:
Change `icarDeviceResource` to inherit from `icarResource`.
Fixes #439


tpekeler (approved):
Looks ok for me, even when changes are marked which seem to be unchanged.

PR #452, opened 2024-04-04
Add enum for milk characteristics type

### LINKED ISSUE #444: Improvements to documentation of icarMilkingVisitEventResource
Further to the discussion on #436 we would like to improve the documentation for `icarMilkingVisitEventResource` as follows:
1. Add a description to `milkingMillkWeight` to clarify that this is the milk weight per the ICAR Guidelines (i.e. a "certified" milk weight).
2. Improve the description of `characteristic` in `icarMilkCharacteristicsType` by creating an enumeration file of the values and updating the description to specify that the values of the enumeration should be used.

cookeac:
Add an enumeration for milk characteristics type.

`icarMilkingVisitEventResource` and `icarMilkCharacteristicsType` are quite old and hard to understand.
Ideally milking characteristics would be managed with a scheme and a published list of characteristics with documented values.

A fallback is to make this an enum. Even there, for a few edge cases, moving from a string to an enum (with a list of valid values) might be a breaking change. Therefore, we have created the enum file and are just referencing it from the description.

Resolves #444 (sort of).


erwinspeybroeck (approved):
approved

Issue #462, opened 2024-04-29
Expand icarAttentionCategoryType.json and icarAttentionCauseType.json to include more types


jdonovan-dml:
The Attention API has the ability to become a 'one-stop-shop' for sharing high level device generated insights.  This could be particularly useful for the likes of sort gates.

Can we add the following enums to the icarAttentionCategoryType.json?
Heat
DeviceIssue
Weight

Can we add the following enums to the icarAttentionCauseType.json?
Heat
LowBattery
Offline
UnderWeight
OverWeight
AtTargetWeight

I understand some of these options are available through more comprehensive API's like the WeightEvent and ReproHeatEvents, but for a simple integration where simply a list of insights is shared, the Attention API could be very useful.

cookeac:
I think we want to take care that organisations don't just use `icarAttentionEventResource` and miss placing important data for animal evaluation such as weights and heats into their own events. However, I do agree that for management purposes it makes sense to add some of these categories and causes to the event. I'm adding the tag for discussion at the next meeting(s).

cookeac:
Discussed 2024-05-08. The extensions above make sense, but where weight, health, and heat (and potentially other) attention triggers occur, do we want to somehow embed all the details of those events (weight recording details, health diagnoses, heat strength and insemination window) into the attentions event? Or link the events? 
Some discussion in #466 (Discussions channel).

cookeac:
Discussed 2024-05-16
We agree with adding these categories and causes, but "heat" should not replace the heat event. Just use this as a device/system alert, not a full heat event.

For icarAttentionCategoryType we want to use "Reproduction" instead of "heat".

PR #470, opened 2024-05-17
Addition of attaention catgories and causes #462

erwinspeybroeck:
https://github.com/adewg/ICAR/issues/462


cookeac (approved):
Approving. The linting checks fail but this is because of previous PRs. I will do a seperate PR to fix those.

PR #471, opened 2024-05-17
#458 Add fedding costs and deliveried weight to consumed feed types

### LINKED ISSUE #458: Add support for feeding costs and feed refusal
Several of our partners would like to have more feed related information:

1, The cost of the ingredient/ration. To be defined if this will be the price per kg/lbs/... or the total cost.
2. The amount of refusal. A pen could be fed 1000 kg of feed, but not all of it could be consumed. Currently we use the weight deliverd to the animals as "feedConsumption". But this actually not what they consumed. Should we substract the refusal from the "feedConsumption"? The refusal is measured as a total, not per ingredient. Some parties are not only interested in what the animals have consumed, but also what is not consumed. Since the refusals also have an impact on the costs

My suggestion: 
- Add a "cost" field to both the entitlement and the consumption. 
- Exclude the refusal amounts from the consumption (proportionally divided)
- Take the refusal into account in the costs of the consumption

ErikUniformAgri:
Resolves #458.

- Added costType and used it to register feeding costs of feed/ration
- Added "deliverd" to be able to distinguish between entitled and actual delivered and being able to calculate the refused feed
- Updated descriptions. These types are used for indivudual animals as well as groups

### DIFF
--- a/types/icarConsumedFeedType.json
+++ b/types/icarConsumedFeedType.json
@@ -1,5 +1,5 @@
 {
-  "description": "gives the consumed feed and the amount the animal was entitled to. Amounts are real weights",
+  "description": "gives the consumed feed and the amount the animal/group was entitled to. Amounts are real weights",
   "type": "object",
   "required": [
     "feedId"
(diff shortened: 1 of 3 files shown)


erwinspeybroeck (approved):
approved

cookeac (approved):
This looks good to me, thanks!

PR #481, opened 2024-06-27
Add extended reason codes

### LINKED ISSUE #454: expand departure reasons
I would like to add 2 reasons:

"extensorleg" - don't know the right English word

"muscleSpasm"

An extensor leg appears from a few months of age. The Achilles tendon is permanently tightened by muscle spasm. This can worsen to such an extent that the hind leg barely touches the ground. It appears as if the affected leg is shorter than the other.

### LINKED ISSUE #455: Review departure reason list
We agreed it would be good to pool our departure reason lists and decide on a common master list.
We may still want to have departure kind (how), departure reason (why - main reason for an animal leaving), and an extended or detailed reason.

Raised during discussion on #454 

Please add your lists as comments to this issue.

cookeac:
Define icarReasonIdentifierType and add its well-known file for people to document their schemes.

Add an array of extended reasons to the death, departure, and do-not-breed events.

Resolves #454 
Resolves #455

### DIFF
--- a/resources/icarMovementDeathEventResource.json
+++ b/resources/icarMovementDeathEventResource.json
@@ -1,5 +1,5 @@
 {
-  "description": "Event for recording animal death",
+  "description": "Event for recording animal death on farm.",
 
   "allOf": [{
       "$ref": "../resources/icarAnimalEventCoreResource.json"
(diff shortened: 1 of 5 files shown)


cookeac:
I propose that we only progress and merge this PR into Develop **after** we have dealt with any minor patch to ADE 1.4 to avoid confusion.

erwinspeybroeck (approved):

PR #483, opened 2024-06-27
Sorting command with multiple sorting sites

### LINKED ISSUE #477: Sorting command with multiple sorting sites
We are currently implementing the new sorting endpoints. However, we encountered a problem with the scenario that there can be multiple sorting sites for the same sorting command.

In some situations, a milking parlour may have two exits, each with its own sorting gate. Each sorting gate has its own sorting sites. In advance, it is not possible to know through which exit the animal will leave the parlour. Therefore, our system supports the option to select multiple sorting sites for the same sorting command. This means one command is issued, and it is considered completed when the animal is sorted at one of the possible sorting sites.

Would it be possible to change the icarAnimalSortingCommandResource to include a list of sorting sites instead of having only one sorting site?

NedapLivestockManagement:
Changed from single sorting site to multiple sortings sites for the animal sorting.

### DIFF
--- a/resources/icarAnimalSortingCommandResource.json
+++ b/resources/icarAnimalSortingCommandResource.json
@@ -16,9 +16,12 @@
           "$ref": "../types/icarAnimalIdentifierType.json",
           "description": "Unique animal scheme and identifier combination."
         },
-        "site": {
-          "type": "string",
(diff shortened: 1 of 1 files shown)


cookeac (approved):
I have reviewed and also made "sites" required.

PR #490, opened 2024-10-23
Feat/add agedin aged out to movement reasons kinds

markusPirkl:


cookeac:
@markusPirkl I think maybe you intended this PR to be in Map of Ag's branch of ICAR ADE?
We would welcome this PR if the only files changed were the two relevant enums, and the PR was accompanied by an ICAR ADE Issue (which I can help with).

PR #492, opened 2024-10-25
Get all animal events

AndreasSchultzGEA:
new Endpoint to get animals complete history (movements and reproduction) at once.


damare:
You don't always want to retrieve all the events of an animal. Moreover all other GET-endpoints define “meta-modified-from” and “meta-modified-to” as optional parameters? Maybe these should also be included here.

If the parameters are omitted during retrieval, they could be given default values. However, these should then be defined in the standard, so that there is a uniform behavior across all partners.

Does "complete history" of an animal include times where the owner was different compared to now? Is it always allowed to change information of an animal of a time that does not overlap with the current ownership?

AndreasSchultzGEA:
Hi damare, this solution is not in the race anymore.

PR #496, opened 2024-11-08
fix nullable reference

### LINKED ISSUE #497: Fix schema problem with nullable reference types (dates) in icarMetaDataType.
Recorded by @damare: 

 There seems to be an issue with nullable $ref-Types in IcarMetaData leading to

`INFO o.o.codegen.InlineModelResolver - allOf schema used by the property created replaced by its only item (a type)`

The fix follows the suggested way discussed here: https://stackoverflow.com/a/48114924

With this fix, the output of the InlineModelResolver disappears.

_Note: I'm just adding the issue here for changelog purposes._

damare:
There seems to be an issue with nullable $ref-Types in IcarMetaData leading to

> INFO o.o.codegen.InlineModelResolver - allOf schema used by the property created replaced by its only item (a type)

The fix follows the suggested way discussed here: https://stackoverflow.com/a/48114924

With this fix, the output of the InlineModelResolver disappears.

### DIFF
--- a/resources/icarAttentionEventResource.json
+++ b/resources/icarAttentionEventResource.json
@@ -12,8 +12,10 @@
 
         "properties": {
             "alertEndDateTime": {
-                "type": ["string", "null"],
-                "$ref": "../types/icarDateTimeType.json",
+                 "anyOf": [
+                     {"type": "null"},
+                     {"$ref": "../types/icarDateTimeType.json"}
+                 ],
                 "description": "RFC3339 date time that represents the end time of an alert (start time is the eventDateTime) if it has ended."
             },
             "category": {
--- a/resources/icarProgenyDetailsResource.json
+++ b/resources/icarProgenyDetailsResource.json
@@ -24,7 +24,11 @@
         "alternativeIdentifiers": {
           "anyOf": [
               {"type": "null"},
-              {"$ref": "../types/icarAnimalIdentifierType.json"}
+              {"type": "array",
+                  "items": {
+                      "$ref": "../types/icarAnimalIdentifierType.json"
+                  }
+              }  
           ],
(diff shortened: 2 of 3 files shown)


cookeac (approved):
This looks like a good change to me. At the time of writing the original we were reflecting what was the best way to allow nullable in combination with a $ref.

mtech-ashelepaev (commented):
For me, these change are valid. Should we also modify the remaining nullable fields within the same PR? icarAttentionEventResource, icarProgenyDetailsResource and etc?

cookeac:
> For me, these change are valid. Should we also modify the remaining nullable fields within the same PR? icarAttentionEventResource, icarProgenyDetailsResource and etc?

Yes we should. It is not every instance of ["string", "null" ] or equivalent (which is valid by itself), but only where a "$ref", object, or array type is used with a null type option. Having reviewed, the only other cases I can find are in these two additional files that Alexey has mentioned.

cookeac:
> > For me, these change are valid. Should we also modify the remaining nullable fields within the same PR? icarAttentionEventResource, icarProgenyDetailsResource and etc?
> 
> Yes we should. It is not every instance of ["string", "null" ] or equivalent (which is valid by itself), but only where a "$ref", object, or array type is used with a null type option. Having reviewed, the only other cases I can find are in these two additional files that Alexey has mentioned.

**Discussed at the meeting on 28 Nov 2024:** 
@damare Daniel - would you be willing to make the equivalent change to `icarAttentionEventResource` and `icarProgenyDetailsResource` as additional commits to this PR, and we will approve these together?

damare:
@cookeac Sure, I will also provide these changes in icarAttentionEventResource and icarProgenyDetailsResource.

cookeac (approved):
This all looks good now, thanks. One more approving review and I'll merge.

mtech-ashelepaev (approved):

AlexeyHardCode (approved):

PR #499, opened 2024-11-25
Issue 493 filter events by animal

AndreasSchultzGEA:


cookeac (commented):
We discussed this at the group meeting Nov 27 2004. We preferred this approach (using animal identity as an optional filter).
- To be consistent with other filters, it would appear that `animal-scheme` and `animal-id` should be the filter parameters. However, I understand that `animal-scheme` is not completely obvious, which is why you have coded it as `animal-identifier-scheme`. Would you consider `animal-id-scheme`?
- We need some documentation so that people know both values must be provided for the filter to work.

AndreasSchultzGEA:
I do have no problems with "animal-id-scheme"

PR #503, opened 2024-12-30
Update icarLocationIdentifierType.md

Busfi3ld:
Introducing uk.mro.herdid to the scheme as a commonly used identifier for UK dairy herds


cookeac (approved):
Excellent, thank you.

PR #514, opened 2025-02-10
Add New Zealand participant codes for location scheme

troygNZ:
Addition of New Zealand participant codes, as commonly used for herd identification.

Using a namespace style approach to allow for extension and consistency as the list grows.
Also being very explicit around style of participant code (dairy + herd), since they have been used in different contexts (enterprises vs individuals vs herds).

Note: Preemptively referring to the upcoming 5 letter participant codes. Can make it 4 for now if preferred.


cookeac:
Agree, looks good.

cookeac (approved):

PR #520, opened 2025-02-13
Add POST methods for icarObservationSummaryResource

### LINKED ISSUE #510: Add POST API method to ObservationSummaryResource
Add a POST so this data can be pushed for exchange as well as fetched.

cookeac:
Add single and batch POST methods for icarObservationSummaryResource.
Resolves #510 

Also corrects one problem with the definition of single POST for icarRemarkResource.


erwinspeybroeck (approved):
ok to me

AlexeyHardCode (approved):

PR #524, opened 2025-02-20
Add Euthanized to icarDeathMethodType

### LINKED ISSUE #515: Add "Euthanized" to icarDeathMethod
### Discussed in https://github.com/adewg/ICAR/discussions/513

<div type='discussions-op-text'>

<sup>Originally posted by **jdoemel** February  1, 2025</sup>
_Could euthanized be added as a deathMethod to the GET and POST deaths endpoints_</div>

I agree, euthanised/euthanized would be a method of death, not a reason.
I'll add this as an issue for discussion with the group. I think most people have used "Perished" which implies died for other than an accident, but doesn't distinguish between animals which died of sickness vs. those that had to be euthanised for sickness.

cookeac:
Add "Euthanized" enum to icarDeathMethodType.
Resolves #515 

As an aside, we had a conversation about "Culled" vs. "Slaughtered" vs. "Euthanized".
Different systems will use this in different ways. "Culled" is often used as a reason for departures (unwanted animals are sold - culled). However in the context of deaths:
* Slaughter - typically means the animal has been killed for food (for people or other animals)
* Culled - may be used when animals are killed to stop disease spreading, or for biosecurity requirements
* Euthanized - describes an animal killed humanely because otherwise it would suffer (e.g from disease or injury)

We also talked about the challenges of documentation for enums, and I noted this conversation in the JSON Schema specification development: https://github.com/json-schema-org/json-schema-spec/issues/57#issuecomment-247855548 (just interesting).

### DIFF
--- a/enums/icarDeathMethodType.json
+++ b/enums/icarDeathMethodType.json
@@ -1,5 +1,5 @@
 { 
-    "description": "Enumeration for method of death (whether the animal died from an accident, natural causes, or was euthanised).",
+    "description": "Enumeration for method of death.\n Note: Some organisations will use Euthanized and Culled as synonyms. Culling may be euthanasia for disease control purposes.\n Culling is also used to describe live animal departures (out of this scope).",
     
     "type": "string",
   
@@ -10,6 +10,7 @@
       "Theft",
       "Lost",
       "Accident",
+      "Euthanized",
       "Other"
     ]
   }


erwinspeybroeck (approved):

cookeac:
Reviewed and approved by the group in both time zones.

PR #525, opened 2025-02-20
Filter events by animal-id

AlexeyHardCode:
animal-scheme, animal-id and animal-identifier parameters have been added to the following methods:
1. Registration scheme: 
        - GET arrivals
	- GET departures
2. Reproduction scheme:
	- GET pregnancy-checks
	- GET heats
	- GET inseminations
	- GET drying-offs
	- GET abortions
	- GET do-not-breeds
	- GET parturitions
	- GET mating-recommendations
	- GET gestations
	- GET repro-status-observations
	- GET repro-embyro-flushings (TYPO in the embryo word, which is corrected in PR)

image

### DIFF
--- a/url-schemes/feedURLscheme.json
+++ b/url-schemes/feedURLscheme.json
@@ -252,6 +252,12 @@
                     },
                     {
                         "$ref": "#/components/parameters/meta-modified-to"
+                    },
+                    {
+                        "$ref": "#/components/parameters/animal-scheme"
+                    },
+                    {
+                        "$ref": "#/components/parameters/animal-id"
                     }
                 ],
                 "responses": {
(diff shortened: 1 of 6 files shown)


AndreasSchultzGEA (changes requested):
why are the new params missing for get-births and get-deaths?

AlexeyHardCode:
> why are the new params missing for get-births and get-deaths?

Very good point, @AndreasSchultzGEA. I was a bit uncertain about whether we should reproduce these parameters for both actions. Normally, an animal has only one birth event and either one or no death event. Do you think we need to fetch the individual birth and death events for the animal? I'm just thinking about use case.

AndreasSchultzGEA (approved):

erwinspeybroeck (approved):

cookeac (approved):

PR #529, opened 2025-03-11
add AnimalTemperatureIncrease and Decrease to attention causes

### LINKED ISSUE #528: Extend causes for Attentions
I would like to extend the `icarAttentionCauseType` with the following two causes:

- AnimalTemperatureIncrease
- AnimalTemperatureDecrease

and for backwards-compatibility keep the `AnimalTemperature` as it is for now.

For our customers, it only makes sense to distinguish between a decrease and increase of the AnimalTemperature, because they have different workflows for both cases, might want to sort only Temperature Increases or sort the cows differently on each case.

I'm not sure if it makes sense to keep the basic `AnimalTemperature` or if we want to remove that one for a future breaking release.
Personally I'd prefer to have it in the standard rather sooner than later.

KrennKristof:
https://github.com/adewg/ICAR/issues/528

### DIFF
--- a/enums/icarAttentionCauseType.json
+++ b/enums/icarAttentionCauseType.json
@@ -6,6 +6,8 @@
     "enum": [
       "Activity",
       "AnimalTemperature",
+      "AnimalTemperatureDecrease",
+      "AnimalTemperatureIncrease",
       "BodyCondition",
       "EatingLess",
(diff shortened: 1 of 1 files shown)


cookeac (approved):
Reviewed with both time zone groups and approved.

PR #545, opened 2025-09-02
539_group_feedings_required

### LINKED ISSUE #539: icarGroupFeedingEventResource missing 'required' specification
The `icarGroupFeedingEventResource` is missing the required specification that the other group events have.
Based on the other events this should be:
` "required" : [ "eventDateTime", "location", "groupMethod" ] `

Adding it now could be a breaking change for some implementations.
Its likely that most implementations would have `eventDateTime` and `location` set. Could we consider adding at least this?

The reason for suggesting adding this is because otherwise malformed JSON still validates as though it was correct.

cookeac:
Add required fields to `icarGroupFeedingEventResource`.
Needs `eventDateTime`, `location`, `groupMethod` to make sense and align with other group events.

Resolves #539

### DIFF
--- a/resources/icarGroupFeedingEventResource.json
+++ b/resources/icarGroupFeedingEventResource.json
@@ -1,6 +1,7 @@
 {
     "description": "Event for recording group or mob feeding. Allowances represent averages so this cannot be used to populate individual animal events.",
     "type": "object",
+    "required": ["eventDateTime", "location", "groupMethod"],
     "allOf": [
         {
             "$ref": "../resources/icarGroupEventCoreResource.json"
--- a/resources/icarGroupFeedingEventResource.json
+++ b/resources/icarGroupFeedingEventResource.json
@@ -1,7 +1,7 @@
 {
     "description": "Event for recording group or mob feeding. Allowances represent averages so this cannot be used to populate individual animal events.",
     "type": "object",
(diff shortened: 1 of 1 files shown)


resources/icarGroupFeedingEventResource.json
```diff
@@ -1,6 +1,7 @@
 {
     "description": "Event for recording group or mob feeding. Allowances represent averages so this cannot be used to populate individual animal events.",
     "type": "object",
+    "required": ["eventDateTime", "location", "groupMethod"],
```
AndreasSchultzGEA: "groupMethod" is also a required property in the parent.
Unsure, if this repetition is needed

AlexeyHardCode: @AndreasSchultzGEA good finding! it seems this "groupMethod" is already required, so, no need to make it reqired in the icarGroupFeedingEventResource.

cookeac: I was also unsure if we needed to add `groupMethod` again when adding the other required properties. If code generation will merge the required lists, then we don't need to add it again. However, it does help with readability to be able to look at the concrete object and see all required fields.

AndreasSchultzGEA (approved):

PR #546, opened 2025-09-02
534_device_animal_simplistic

### LINKED ISSUE #534: Cater for on-animal devices by adding animal identifier to icarDeviceResource
The `icarDeviceResource` is designed to be used where device information is needed, but also to be potentially enumerable for a location (here are the devices on this farm).

Some of these devices might be on-animal (or in-animal) devices. It would be useful to optionally include the ID of the animal in the device resource.

cookeac:
Simplest solution to #534.
Adds optional `animal` (`icarAimalIdentifierType`) to `icarDeviceResource` to provide the currently associated animal (if any). 

Comments welcome:
* Is the name sufficient, or would `associatedAnimal` be better? Most other animal identifiers are just called `animal` where used.
* We discussed an array of identifiers with validity periods to provide a historic view, but most people felt this would be confusing and unnecessary.
* We discussed join/leave or `association` events, but I have left this out of initial scope.

Resolves #534

### DIFF
--- a/resources/icarDeviceResource.json
+++ b/resources/icarDeviceResource.json
@@ -59,7 +59,11 @@
       "registration": {
         "$ref": "../types/icarDeviceRegistrationIdentifierType.json",
         "description": "A registration identifier for the device (most devices should eventually have a registration issued by `org.icar` or other entity)."
-      }
+      },
+        "animal": {
+          "$ref": "../types/icarAnimalIdentifierType.json",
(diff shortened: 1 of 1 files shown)


cookeac:
@andre-wilson40 and @krennkristof for your comments

AndreasSchultzGEA (approved):

mtech-ashelepaev (approved):

AlexeyHardCode (approved):

PR #562, opened 2025-10-22
Wrap $ref with allOf for JSON snippets

cookeac:
With OpenAPI 3.1, in theory external JSON files that don't contain a schema declaration need to be wrapped in "allOf". Some linters are now enforcing this. Added "allOf" around external $refs.

### DIFF
--- a/url-schemes/healthURLScheme.json
+++ b/url-schemes/healthURLScheme.json
@@ -1240,7 +1240,7 @@
       "batchResults": {
         "type": "array",
         "items": {
-            "$ref": "../resources/icarBatchResult.json" 
+          "$ref": "../resources/icarBatchResult.json"
         }
       },
       "icarDiagnosisEvent": {
--- a/url-schemes/healthURLScheme.json
+++ b/url-schemes/healthURLScheme.json
@@ -1240,9 +1240,7 @@
       "batchResults": {
         "type": "array",
         "items": {
-          "allOf": [
-            { "$ref": "../resources/icarBatchResult.json" }
-          ]
+            "$ref": "../resources/icarBatchResult.json" 
         }
       },
       "icarDiagnosisEvent": {
@@ -1258,9 +1256,7 @@
       "icarDiagnosisEventArray": {
         "type": "array",
         "items": {
-          "allOf": [
-            { "$ref": "../resources/icarDiagnosisEventResource.json" }
-          ]
+             "$ref": "../resources/icarDiagnosisEventResource.json" 
         }
       },
       "icarTreatmentEvent": {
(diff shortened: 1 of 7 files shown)


url-schemes/healthURLScheme.json
```diff
@@ -1240,74 +1240,98 @@
       "batchResults": {
         "type": "array",
         "items": {
-          "$ref": "../resources/icarBatchResult.json"
+            "$ref": "../resources/icarBatchResult.json"
```
AndreasSchultzGEA: aren't this 2 additional spaces wrong?

url-schemes/managementURLScheme.json
```diff
@@ -1897,14 +1897,20 @@
             "batchResults": {
```
AndreasSchultzGEA, line 1897: why does this batchResults-definition differ from the one in the healthManagementURLScheme-definition?
There, a single $ref is used instead of a allOf at this place?

AndreasSchultzGEA (commented):
Hi Andrew, I could not be a help on this review, as I have no knowledge about the underlying problem, nore are the rules to fix this problem obvious.

AlexeyHardCode:
I'll do testing with code generation and report about results.

AlexeyHardCode:
@cookeac Here is my tests results. 

Schema bundling completed with warnings.
image
This occurs because the allOf construct creates a new composed schema object with the same name around the referenced schema.As a result, the bundled schema contains two components — icarDiagnosisEventCollection and icarDiagnosisEventCollection-2 — where the first simply references the second.
image
Additionally, when generating the code, it creates many similar models. In my opinion, this adds unnecessary complexity and makes the codebase harder to maintain.
image

cookeac:
> This occurs because the allOf construct creates a new composed schema object with the same name around the referenced schema.As a result, the bundled schema contains two components — icarDiagnosisEventCollection and icarDiagnosisEventCollection-2 — where the first simply references the second.

@AlexeyHardCode is this specific to bundling and code generation from the bundled schema, or does it also occur with the individual OpenAPI files? Obviously if it occurs in all instances we should roll back this change -- but the tradeoff will be many linting errors because of non-compliance with the 3.1 specs. Happy to do this for now, but recognising this will be a maintenance issue in future.

AlexeyHardCode:
@cookeac I think allOf will always create a new component, while $ref simply references an existing schema. Could you please share the linters list that enforce the use of allOf?
In VS Code, I’ve also noticed a warning message stating ‘Property $ref is not allowed.’ I’m thinkg how to improve validation rules to avoid such warnings.
image

cookeac:
Closing this pull request without merging. We will investigate the linting errors further and see if there is another solution OR if a future version of the bundling script can be adjusted to handle "allOf" (it does not at present).

PR #579, opened 2025-11-27
Refactor AnimalCore and ProgenyDetails recources. new AnimalBase resource.

### LINKED ISSUE #564: Add parents and other attributes in parturition icarProgenyDetailsResource
Why does the icarProgenyDetailsResource does not contain any possibility to provide parents?
Somehow, I missed the decision on this.

AlexeyHardCode:
Implementing the Option6 for the Parturition event. 
Option 6 - Refactor icarAnimalCoreResource and icarProgenyDetailsResource so they share the same fields, but icarProgenyDetailsResource has optional identifier and icarAnimalCoreResource the identifier is required.

### DIFF
--- a/resources/icarAnimalBaseResource.json
+++ b/resources/icarAnimalBaseResource.json
@@ -71,10 +71,6 @@
             "$ref": "../types/icarParentageType.json"
           }
         },
-        "status": {
-          "$ref": "../enums/icarAnimalStatusType.json",
-          "description": "On-farm status of the animal (such as alive, dead, off-farm)."
-        },
         "healthStatus": {
           "$ref": "../enums/icarAnimalHealthStatusType.json",
           "description": "Health status of the animal (such as Healthy, Suspicious, Ill, InTreatment, ToBeCulled)."
--- a/resources/icarAnimalBaseResource.json
+++ b/resources/icarAnimalBaseResource.json
@@ -1,0 +1,85 @@
+{
+  "description": "Base schema containing all common properties for animal and progeny resources.",
+
+  "allOf": [{
+      "$ref": "../resources/icarResource.json"
+    },
+    {
+      "type": "object",
+
+      "required": [
+        "specie",
+        "gender"
+      ],
+      
+      "properties": {        
+        "alternativeIdentifiers": {
+          "type": "array",
+          "items": {
+            "$ref": "../types/icarAnimalIdentifierType.json"
+          },
+          "description": "Alternative identifiers for the animal. Here, also temporary identifiers, e.g. transponders or animal numbers, can be listed."
+        },
+        "specie": {
+          "$ref": "../enums/icarAnimalSpecieType.json",
+          "description": "Species of the animal."
+        },
+        "gender": {
+          "$ref": "../enums/icarAnimalGenderType.json",
+          "description": "Gender of the animal."
+        },
+        "birthDate": {
+          "$ref": "../types/icarDateTimeType.json",
+          "description": "RFC3339 UTC date/time of birth (see https://ijmacd.github.io/rfc3339-iso8601/ for how to use)."
+        },
+        "primaryBreed": {
+          "$ref": "../types/icarBreedIdentifierType.json",
+          "description": "ICAR Breed code for the animal."
+        },
+        "breedFractions": {
+          "$ref": "../types/icarBreedFractionsType.json",
+          "description": "Breed fractions for the animal."
+        },
+        "coatColor": {
+          "type": "string",
+          "description": "Colour of the animal's coat, using the conventions for that breed."
+        },
+        "coatColorIdentifier": {
+          "$ref": "../types/icarCoatColorIdentifierType.json",
+          "description": "Colour of the animal's coat using a national or breed-defined scheme and identifier combination."
+        },
+        "managementTag": {
+          "type": "string",
+          "description": "The identifier used by the farmer in day to day operations. In many cases this could be the animal number."
+        },
+        "name": {
+          "type": "string",
+          "description": "Name given by the farmer for this animal."
+        },
+        "officialName": {
+          "type": "string",
+          "description": "Official herdbook name."
+        },
+        "productionPurpose": {
+          "$ref": "../enums/icarProductionPurposeType.json",
+          "description": "Primary production purpose for which animals are bred."
+        },
+        "reproductionStatus": {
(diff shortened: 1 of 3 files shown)


AlexeyHardCode:
I had to add Identifier property into both icarAnimalCoreResource, icarProgenyDetailsResource files, because code generation doesn't work well with required fileds for inherited properties. I also found, that we have few places in the other resources where these required operator doesn't apply, I'll report about those in additional ISSUE.

tpekeler (changes requested):
These Changes worked for our generation.
Two comments:

1. Would it be possible to also move the identifier field to the base class?
edit: (resolved, as mentioned by Alexey above, didn't recognized before)
2. Why isn't the status in the base class, but the reproductionStatus?

I totally understand, that a lactationStatus is only suitable for adult cows, so I would assume the same goes for reproductionStatus. Submitting a dead pogeny would be beneficial, I suppose.

cookeac:
> 2. Why isn't the status in the base class, but the reproductionStatus?
> I totally understand, that a lactationStatus is only suitable for adult cows, so I would assume the same goes for reproductionStatus. Submitting a dead pogeny would be beneficial, I suppose.

I agree, I suggest moving reproductionStatus to icarAnimalCoreResource and status into icarAnimalBaseResource.

cookeac:
This looks good. I wondered if we should put `icarAnimalBaseResource` into the `types` folder and referencing it from there as it is not a concrete resource that you can fetch. I believe this would reduce warnings for pydantic generation (@jrobichaud may comment on this).

tpekeler:
> This looks good. I wondered if we should put `icarAnimalBaseResource` into the `types` folder and referencing it from there as it is not a concrete resource that you can fetch. I believe this would reduce warnings for pydantic generation (@jrobichaud may comment on this).

Don't know about all the theory, but  icarResource is also defined inside resources.

AndreasSchultzGEA:
no additional comments.
Inheritance seems to be okay.

AlexeyHardCode:
> > 2. Why isn't the status in the base class, but the reproductionStatus?
> >    I totally understand, that a lactationStatus is only suitable for adult cows, so I would assume the same goes for reproductionStatus. Submitting a dead pogeny would be beneficial, I suppose.
> 
> I agree, I suggest moving reproductionStatus to icarAnimalCoreResource and status into icarAnimalBaseResource.

@tpekeler @cookeac 
I thought about theese both Statuses during the implementation. My idea was that Status represents farm-management state (alive, dead, off-farm), which only applies to a full animal record, not to general biological item like progeny. For progeny, birthStatus is sufficient to report whether the progeny was born alive, dead, or stillborn, so they don’t need the full status field from the animal resource. Do you think we would rather have both statuses in the Base model?
I agree that reproductionStatus must belong only to AnimalCoreResource. Moved it. 
@cookeac Good point, Andrew. I will rename icarAnimalBaseResource to icarAnimalBaseType and move it to types. This will better indicate that it is a base resource that shouldn’t be used directly.

tpekeler:
> @tpekeler @cookeac I thought about theese both Statuses during the implementation. My idea was that Status represents farm-management state (alive, dead, off-farm), which only applies to a full animal record, not to general biological item like progeny. For progeny, birthStatus is sufficient to report whether the progeny was born alive, dead, or stillborn, so they don’t need the full status field from the animal resource. Do you think we would rather have both statuses in the Base model? I agree that reproductionStatus must belong only to AnimalCoreResource. Moved it. @cookeac Good point, Andrew. I will rename icarAnimalBaseResource to icarAnimalBaseType and move it to types. This will better indicate that it is a base resource that shouldn’t be used directly.

Didn't see the BirthStatus in the comparison of the files. Therefore I think all 3 status Attributes should be left in icarAnimalCoreResource. 
Regarding Type vs Resource we should come to a common rule of how to use what. Don't know whether moving the Base is resulting in additional issues. Maybe thats a task for more wider refactoring to implement this rule.

PS closed this PR accidently, hope nothing went wrong

AlexeyHardCode:
Status moved to AnimalBase resource while reproductionStatus to AnimalCore resource (https://github.com/adewg/ICAR/pull/579/commits/c64f42e5fc1e1923d7d2bf1a45823b6b735c00d0) No changes to Type vs Resource and file relocation.

tpekeler:
> Status moved to AnimalBase resource while reproductionStatus to AnimalCore resource ([c64f42e](https://github.com/adewg/ICAR/commit/c64f42e5fc1e1923d7d2bf1a45823b6b735c00d0)) No changes to Type vs Resource and file relocation.

Hi, I think you were right about the status in the first place, like I mentioned above, I didn't recognized the birthStatus in the PogenyDetails. Therefore I think the status in the base would be somhow duplikate information. Like mentioned I think it would be best to have status, lactationStatus and reproductionStatus for the CoreAnimal and birthStatus for the Pogeny. The base should only contain everything suitable for both.

AlexeyHardCode:
@tpekeler I agree.  
1. **AnimalBase**: healthStatus, productionPurpose, parentage, alternativeIdentifiers and other more bovious fields
2. **CoreAnimal**: identifier (required), status, lactationStatus and reproductionStatus.
3. **ProgenyDetails**: identifier (optional), taggingDate, birthStatus, birthSize, birthWeight.

@cookeac do you also agree?

cookeac:
> 1. **AnimalBase**: healthStatus, productionPurpose, parentage, alternativeIdentifiers and other more bovious fields
> 2. **CoreAnimal**: identifier (required), status, lactationStatus and reproductionStatus.
> 3. **ProgenyDetails**: identifier (optional), taggingDate, birthStatus, birthSize, birthWeight.

Yes, that makes sense. I had forgotten the apparent duplication between `birthStatus` and `status`.

tpekeler (approved):
This seems to be right now.

AlexeyHardCode:
@cookeac, @erwinspeybroeck @AndreasSchultzGEA Could you please review and approve this PR so that we can include it in the next ADE release? On our side, we have tested code generation with C# and Java, verified that the properties are logically distributed between the animalCore and progenyDetails objects, and ensured that the changes do not introduce any breaking changes.
We are in a bit of a hurry with these updates, which is why we are hoping to get it approved as soon as possible. Please accept our apologies for the urgency.

AndreasSchultzGEA (approved):

erwinspeybroeck (approved):

cookeac (approved):

PR #581, opened 2025-11-27
Extension of ProductionPurposeType.

### LINKED ISSUE #566: Add further enumerated values to icarProductionPurposeType
In issue #564 we identified that it might be useful to have additional values in `icarProductionPurposeType`.

Currently this only has enumerated values "meat", "milk", "wool" and the `productionPurpose` in `icarAnimalCoreResource` is a single value, so you cannot have multiple values for an animal.

Suggestions are:
* Suckler - meaning intended to be sold for meat, even if from milk genetics (may be covered by meat purpose)
* Breeding - EITHER means "intended to be retained for use as a parent" or "used in a stud breeding operation to produce semen or embryos" (we would need to clarify which)
- Test - not sure what this means.

We noted that these are potentially a separate set of values to the meat/milk/wool product purposes - more of a management purpose than a production purpose? Although semen / embryos might be valid production purposes.

We agreed to investigate further.

AlexeyHardCode:
Added Suckler, Breeding, TestAnimal, Pet options to the ProductionPurposeType.

### DIFF
--- a/enums/icarProductionPurposeType.json
+++ b/enums/icarProductionPurposeType.json
@@ -1,11 +1,15 @@
 {
-  "description": "Defines the primary product that for which this animal is bred or kept.\\nIf animals are kept for breeding or live trade (sale), specify the end purpose of that breeding/trade.\\n\n - Meat corresponds to UNSPC 50111500 Minimally processed meat and poultry products\\n - Milk corresponds to UNSPC 50203200 Raw milk products from live animals\\n - Wool corresponds to UNSPC 11131506 Unprocessed wool",
+  "description": "Defines the primary product that for which this animal is bred or kept.\\nIf animals are kept for breeding or live trade (sale), specify the end purpose of that breeding/trade.\\n\n - Meat corresponds to UNSPC 50111500 Minimally processed meat and poultry products\\n - Milk corresponds to UNSPC 50203200 Raw milk products from live animals\\n - Wool corresponds to UNSPC 11131506 Unprocessed wool.\\n Suckler - used to breastfeed calves.\\n Breeding - bulls used for semen production.\\n TestAnimal - animals used for research.\\n",
 
   "type": "string",
 
   "enum": [
     "Meat",
     "Milk",
-    "Wool"
+    "Wool",
+    "Suckler",
+    "Breeding",
+    "TestAnimal",
+    "Pet"
   ]
 }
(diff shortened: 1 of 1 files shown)


enums/icarProductionPurposeType.json
```diff
@@ -1,11 +1,15 @@
...
   "type": "string",
 
   "enum": [
     "Meat",
     "Milk",
-    "Wool"
+    "Wool",
+    "Suckler",
+    "Breeding",
+    "TestAnimal",
```
cookeac:
```suggestion
    "Research",
```

cookeac: Thinking about it, maybe "Research" is clearer.

enums/icarProductionPurposeType.json
```diff
@@ -1,11 +1,15 @@
 {
-  "description": "Defines the primary product that for which this animal is bred or kept.\\nIf animals are kept for breeding or live trade (sale), specify the end purpose of that breeding/trade.\\n\n - Meat corresponds to UNSPC 50111500 Minimally processed meat and poultry products\\n - Milk corresponds to UNSPC 50203200 Raw milk products from live animals\\n - Wool corresponds to UNSPC 11131506 Unprocessed wool",
+  "description": "Defines the primary product that for which this animal is bred or kept.\\nIf animals are kept for breeding or live trade (sale), specify the end purpose of that breeding/trade.\\n\n - Meat corresponds to UNSPC 50111500 Minimally processed meat and poultry products\\n - Milk corresponds to UNSPC 50203200 Raw milk products from live animals\\n - Wool corresponds to UNSPC 11131506 Unprocessed wool.\\n Suckler - used to breastfeed calves.\\n Breeding - bulls used for semen production.\\n TestAnimal - animals used for research.\\n",
```
cookeac:
```suggestion
  "description": "Defines the primary product that for which this animal is bred or kept.\\nIf animals are kept for breeding or live trade (sale), specify the end purpose of that breeding/trade.\\n\n - Meat corresponds to UNSPC 50111500 Minimally processed meat and poultry products\\n - Milk corresponds to UNSPC 50203200 Raw milk products from live animals\\n - Wool corresponds to UNSPC 11131506 Unprocessed wool.\\n Suckler - used to breastfeed calves.\\n Breeding - bulls used for semen production.\\n Research - animals used for research.\\n",
```

cookeac (changes requested):

enums/icarProductionPurposeType.json
AlexeyHardCode: I think it's a good idea. Much clearer. I'll change.

enums/icarProductionPurposeType.json
AlexeyHardCode: Perhaps we don't need exmpalantion for Research option is this case. It's more than clear.

cookeac (approved):

PR #587, opened 2026-01-23
Sire primary breed in insemination

AndreasSchultzGEA:


erwinspeybroeck (approved):

cookeac (approved):

AlexeyHardCode:
@AndreasSchultzGEA overall looks very good. One suggestion I would like like to propose. 
These three schemas contain a clearly duplicated set of sire-related properties: **sireIdentifiers**, **sireOfficialName**, **sireURI**, **sirePrimaryBreed**. Extracting them into a shared reusable schema **icarReproSireInfoResource** reduces duplication and provides a single point of maintenance. This refactoring is not a breaking change and does not impact ADE version upgrades, as it preserves the existing data model while only reorganizing the schema structure.
@cookeac @erwinspeybroeck do you also agree? I can quickly refactor these three resources and upload a new version for you to check. 

The same three properties (excluding **sirePrimaryBreed**) are also present in icaSireRecommendationType within icarReproMatingRecommendationResource.
Would it be appropriate to include **sirePrimaryBreed** in icaSireRecommendationType as well?

cookeac:
> One suggestion I would like like to propose. These three schemas contain a clearly duplicated set of sire-related properties: **sireIdentifiers**, **sireOfficialName**, **sireURI**, **sirePrimaryBreed**. Extracting them into a shared reusable schema **icarReproSireInfoResource** reduces duplication and provides a single point of maintenance. This refactoring is not a breaking change and does not impact ADE version upgrades, as it preserves the existing data model while only reorganizing the schema structure. @cookeac @erwinspeybroeck do you also agree? I can quickly refactor these three resources and upload a new version for you to check.
> 
> The same three properties (excluding **sirePrimaryBreed**) are also present in icaSireRecommendationType within icarReproMatingRecommendationResource. Would it be appropriate to include **sirePrimaryBreed** in icaSireRecommendationType as well?

@AlexeyHardCode I had wondered about the duplication as well and I think you have come up with a good solution. We should check that it does not affect code generation, but I am very happy with that approach, and would apply it to `icarSireRecommendationType` as well.

erwinspeybroeck:
Good idea to do so. I agree with both of you ;-)

AlexeyHardCode (approved):

PR #598, opened 2026-03-19
Refactor of the Resource/Types references

### LINKED ISSUE #543: [ICAR-pydantic] Circular reference between types and resources because of icarInventoryTransactionType
I detected a circular dependency with `icarInventoryTransactionType`:

https://github.com/adewg/ICAR/blob/990e3c02bf11ead41550420e2f134c1ac38ee30d/types/icarInventoryTransactionType.json#L6

The type `icarInventoryTransactionType` refers to a resource but it is not supposed to happen.

ICAR Dependency diagram:
```mermaid
flowchart LR
    enums
    resources

    collections
    types
    resources --> types
    resources --> enums
    types --> enums
    collections --> resources
    types --> |'types' should not depend on 'resources'| resources

    linkStyle 4 stroke: red,color: red
```

I believe it should be a resource instead.

This affects the `ICAR-pydantic` when generating the modules since it creates circular imports (when a module import another one and vice versa). (in the meantime, I will assume it is a resource as a workaround)

mtech-ashelepaev:
Move icarEventCoreResource referece from the type to each Transaction Resource

### DIFF
--- a/resources/icarFeedTransactionResource.json
+++ b/resources/icarFeedTransactionResource.json
@@ -7,6 +7,9 @@
             "$ref": "../types/icarInventoryTransactionType.json"
         },
         {
+            "$ref": "../resources/icarEventCoreResource.json"
+        },
+        {
             "type": "object",
 
             "required": [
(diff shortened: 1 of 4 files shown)


cookeac (approved):
This is a good solution. It may cause some code generation changes for some users (e.g. my business) but this can easily be addressed.

erwinspeybroeck (approved):

AndreasSchultzGEA (approved):

Issue #632, opened 2026-07-15
Add device cause types for attentions


NedapLivestockManagement:
I would like to propose two changes to the Attention API.
 
Please add the following values to `icarAttentionCauseType.json`:
 
| Proposed value | Description | 
| --- | --- | 
| `AttachedIncorrectly` | The tag is not attached correctly. | 
| `InvalidPosition` | The tag is attached, but positioned incorrectly.  | 
| `BackToFront` | The tag is attached in the wrong direction.  | 
| `NeverSentData` | The tag has not sent any data since it was activated.| 
| `StoppedSendingData` | The tag was active but has not sent data for a prolonged period. | 
| `InsufficientData` | The tag sends too little data, it might be broken or not in use. | 
| `NoAnimalLinked` | Data is being received from a tag that is not linked to any animal. | 
 
Currently `icarAttentionEventResource.json` extends `icarAnimalEventCoreResource.json`, which requires an `animal`. For device-related attention events this is not always available. For example, `NoAnimalLinked` is specifically a device issue that occurs when no animal is linked to the device. Therefore please make the `animal` field optional in `icarAttentionEventResource.json`.
 
In those cases the attention event should still be publishable with a `device`, `category`, and `causes`, but without an `animal`.

cookeac:
Possibilities:
1. New DeviceAttention event, for the NoAnimalLinked (and other equipment problems).
2. Make animal optional - but this could break existing implementations if an event arrives without an animal.
3. Pass a scheme and the device identifier in the animal - but this puts a device into an animal field, which will also cause confusion. Although once attached to an animal, a device identifier could be a valid animal identifier.

Option 1 is probably the most elegant and would allow expansion to other devices that are not animal wearables.
But we may find that device specific attentions migrated to that new event.

PR #634, opened 2026-07-29
Add device cause types for attentions

NedapLivestockManagement:
## Summary
Adds the new attention cause types from issue #632 and makes `animal` optional on `icarAttentionEventResource` so device-related attention events can be published without a linked animal.

## Changes
- Added the new values to `enums/icarAttentionCauseType.json`
- Made `animal` optional in `resources/icarAttentionEventResource.json`
- Updated attention-event descriptions in the OpenAPI URL schemes
- Added a changelog entry under v1.5.1

## Notes
This keeps attention events publishable with device, category, and causes even when no animal is linked.


resources/icarAttentionEventResource.json
```diff
@@ -1,16 +1,20 @@
...
-        "$ref": "../resources/icarAnimalEventCoreResource.json"
+        "$ref": "../resources/icarEventCoreResource.json"
       },
       {
         "type": "object",
 
         "required" : ["category", "causes"],
 
         "properties": {
+            "animal": {
```
cookeac: This is almost certain to break expectations of recipients that they can resolve the animal to an animal. Having a previous animal event that might no longer contain an animal identifier is a breaking change. We are better to define a separate (but similar) device event that can be used for device or equipment status independent of animals.

cookeac:
Generally my feedback is that repurposing an animal event that has previously been based on `icarAnimalEventCoreResource` and will be expected to have an animal identifier is a breaking change. A better approach would be to define an `icarDeviceEventCoreResource` that expects to have a device identifier or reference, and then use this for device messages (such as NoAnimalLinked) and potentially could also be used for device statistics independent of animals.

cookeac (changes requested):

cookeac:
For this PR, we will also need to adjust the target from adewg:ADE-1 to adewg:Develop

erwinspeybroeck:
There is of course an indication that the animal might not be known --> attentionCategoryType= DeviceIssue
But I understand the reluctance to do so

cookeac (approved):

PR #637, opened 2026-08-11
Add optional death date and disposal method to progeny details

### LINKED ISSUE #611: Adding optional DisposalMethod and DeathDate to ProgenyDetails
The Finnish official register requires disposal method and death date for all animals that have died after having been born alive, including calves that died or were slaughtered or euthanised before tagging. As they are not tagged and so don't have an identifier, the death can't be reported using the death event and needs to be reported in the parturition event. For twins the data might be different, so it should be given by progeny. So I suggest adding optional fields disposalMethod (same as in icarMovementDeathEventResource) and deathDate to icarProgenyDetailsResource.

If other countries that use the birth method instead need the same fields also, they could be either added there also or straight to icarAnimalCoreResource, which is used by both. In icarAnimalCoreResource they might also be usefull for somebody when getting animal data.

mtech-ssipila:
Added optional death date and disposal method fields to progeny details, to be used when progeny has died before tagging.
Closes: adewg/ICAR#611

### DIFF
--- a/resources/icarProgenyDetailsResource.json
+++ b/resources/icarProgenyDetailsResource.json
@@ -40,6 +40,20 @@
               {"type": "null"},
               {"$ref": "../types/icarMassMeasureType.json"}
           ]
+        },
+        "deathDate": {
+          "description": "Progeny death date in RFC3339 UTC (see https://ijmacd.github.io/rfc3339-iso8601/ for format guidance).",
+          "anyOf": [
+              {"type": "null"},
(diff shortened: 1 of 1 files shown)


cookeac:
Looking good. We need to decide if we want these to be nullable anyOf allowing null, or whether they are just optional fields (expected to have a value if specified, but do not have to be specified). These are slightly different semantics in many cases (especially for enumerations).

AndreasSchultzGEA (approved):

PR #638, opened 2026-08-31
Add X-ADE-Version and X-ADE-AcceptVersion headers and tidy OpenAPI specs

### LINKED ISSUE #631: How ADE message version should be identified
ADE has multiple minor versions, and in practice different partners may use different minor ADE versions at the same time, for example ADE 1.3 and ADE 1.5. Some partners would like to know which ADE version was used to produce a specific payload.

Possible options:
**1. Expose the ADE version in an HTTP header**
_Add the ADE version as HTTP metadata, for example in a header._
**2. Include ADE version metadata in the ADE message**
_Add a standard metadata element to the ADE message structure._ 
**3. Add a discovery endpoint for supported ADE specifications**
_Add an endpoint that returns the supported ADE specification version or versions, for example ADE version, schema date, and OpenAPI schema location._

Which approach is preferred or whether another option should be used?

cookeac:
Add the `X-ADE-Version` header to all responses in all OpenAPI scheme files (in `/url-schemes`).
Add the optional `X-ADE-AcceptVersion` header to all request parameters in the OpenAPI files.

Along the way, I noticed a couple of errors (an unused `icarFeedInventoryCollection` in the management schemes, an incorrect comment on `milking-visit` in the milking url schemes so corrected these.

There was a lot of duplication of batch error responses, 201 created and 202 accepted responses, so where I could I added these definitions to `#components/responses` and simplified the individual endpoint definitions.

I know this will be hard to review because of the size of the OpenAPI files, so I will also ask CoPilot to review.

### DIFF
--- a/url-schemes/sortingURLScheme.json
+++ b/url-schemes/sortingURLScheme.json
@@ -36,11 +36,19 @@
           },
           {
             "$ref": "#/components/parameters/location-id"
+          },
+          {
+            "$ref": "#/components/parameters/X-ADE-AcceptVersion"
           }
         ],
         "responses": {
           "200": {
             "description": "Successful. The response contains the resources for the given location.",
+            "headers": {
+              "X-ADE-Version": {
+                "$ref": "#/components/headers/X-ADE-Version"
+              }
+            },
             "content": {
               "application/json": {
                 "schema": {
@@ -75,11 +83,19 @@
           },
           {
             "$ref": "#/components/parameters/animal-id"
+          },
+          {
+            "$ref": "#/components/parameters/X-ADE-AcceptVersion"
           }
         ],
         "responses": {
           "200": {
             "description": "Successful. The response contains the resources for the given location.",
+            "headers": {
+              "X-ADE-Version": {
+                "$ref": "#/components/headers/X-ADE-Version"
+              }
+            },
             "content": {
               "application/json": {
                 "schema": {
@@ -112,11 +128,19 @@
           },
           {
             "$ref": "#/components/parameters/animal-id"
+          },
+          {
+            "$ref": "#/components/parameters/X-ADE-AcceptVersion"
           }
         ],
         "responses": {
           "200": {
-            "description": "Successful."
+            "description": "Successful.",
+            "headers": {
+              "X-ADE-Version": {
+                "$ref": "#/components/headers/X-ADE-Version"
+              }
+            }
           },
           "default": {
             "$ref": "#/components/responses/default"
@@ -138,11 +162,19 @@
           },
           {
             "$ref": "#/components/parameters/location-id"
+          },
+          {
+            "$ref": "#/components/parameters/X-ADE-AcceptVersion"
           }
         ],
         "responses": {
           "200": {
             "description": "Successful. The response contains the resources for the given location.",
+            "headers": {
+              "X-ADE-Version": {
+                "$ref": "#/components/headers/X-ADE-Version"
+              }
+            },
             "content": {
               "application/json": {
                 "schema": {
@@ -169,6 +201,9 @@
           },
           {
             "$ref": "#/components/parameters/location-id"
+          },
+          {
+            "$ref": "#/components/parameters/X-ADE-AcceptVersion"
           }
         ],
         "requestBody": {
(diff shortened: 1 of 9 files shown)


url-schemes/sortingURLScheme.json
```diff
@@ -275,11 +329,25 @@
         "schema": {
           "type": "string"
         }
+      },
+      "X-ADE-AcceptVersion": {
```
AndreasSchultzGEA, line 333: I can't remember, that we have been talking about an accept-version.
What happens, if the requested services can't provide the accept-version? How is this communicated to the client? By an error? By ignoring the accept-version?

cookeac: Having an X-ADE-AcceptVersion does allow a server that supports multiple versions to change its behaviour. This is helpful. 

But indeed - what happens if the X-ADE-AcceptVersion does not agree with what the server handles.
If the major version is different, the server should almost certainly error.
Otherwise, you would expect that with backward compatibility and fault tolerant design, the interoperability should be able to proceed. A client should probably tell its user if the versions do not agree. The server could log the issue for later operator inspection. Per the discussion in #631 it would probably be best if the server only errors if additional data or enum values from the client would cause data loss if ignored.

AlexeyHardCode:
@cookeac thanks for a PR! 

Here is a result of a cosde review:
If I see it right, `X-ADE-AcceptVersion` is missing in a few places:
1. exampleUrlScheme.json - GET - /locations/{location-scheme}/{location-id}/animals
2. exampleUrlScheme.json - GET - /service-sires
3. healthURLScheme.json - POST - /locations/{location-scheme}/{location-id}/diagnoses
4. managementURLScheme.json - POST - /locations/{location-scheme}/{location-id}/animal-set-leaves

Also I would like to suggest a minor schema refactoring. 
As we add more parameters, we could move shared ones (`X-ADE-AcceptVersion`, `location-scheme`, `location-id` etc.) to the path level instead of repeating them in every GET/POST operation. This reduces duplication in the schema, with no change to actual API behavior, parameters and their semantics remain identical for all requests/responses.
Example: 
```
"/locations/{location-scheme}/{location-id}/animal-sets": {
  "parameters": [
    { "$ref": "#/components/parameters/location-scheme" },
    { "$ref": "#/components/parameters/location-id" },
    { "$ref": "#/components/parameters/X-ADE-AcceptVersion" }
  ],
  "get": {
    "parameters": [
      { "$ref": "#/components/parameters/meta-modified-from" },
      { "$ref": "#/components/parameters/meta-modified-to" }
    ],
    ...
  },
  "post": { ... }  // no parameters array needed at all
}
```
Andrew, if you wish, I can help you with implementing and testing that.

cookeac:
> @cookeac thanks for a PR!
> 
> Here is a result of a cosde review: If I see it right, `X-ADE-AcceptVersion` is missing in a few places:
> 
> 1. exampleUrlScheme.json - GET - /locations/{location-scheme}/{location-id}/animals
> 2. exampleUrlScheme.json - GET - /service-sires
> 3. healthURLScheme.json - POST - /locations/{location-scheme}/{location-id}/diagnoses
> 4. managementURLScheme.json - POST - /locations/{location-scheme}/{location-id}/animal-set-leaves

Thanks @AlexeyHardCode - I'll follow up those ones.

> Also I would like to suggest a minor schema refactoring. As we add more parameters, we could move shared ones (`X-ADE-AcceptVersion`, `location-scheme`, `location-id` etc.) to the path level instead of repeating them in every GET/POST operation. This reduces duplication in the schema, with no change to actual API behavior, parameters and their semantics remain identical for all requests/responses. 
> Andrew, if you wish, I can help you with implementing and testing that.

I had been thinking the same thing! But I think we're better to do a new issue and a new PR for that.

AlexeyHardCode:
@cookeac  thanks for a quick response, I agree, better not to mix this. One more thing we might want to add to this refactoring task:
1. sortingURLScheme.json - POST - /locations/{location-scheme}/{location-id}/sorting for some reason only this repsonse explicitly lists  "400"  and  "500" codes. May be default will be just enough.

Otherwise, it looks good. If you don't mind, during today/tomorrow, I'll generate the code and create a JSON spec from it to check that everything is consistent with the previous version. Then inform you.

AlexeyHardCode:
@cookeac Andrew, we have a small problem when combining our schemas into one file, the process crashes with an error saying it can't find  `X-ADE-Version`.
The tool we use for merging (openapi-merge-cli) gets confused by a ref link to a shared header and crashes when comparing files across schemas. This is a known limitation of the tool, not something we did wrong. It can only resolve links pointing to components/schemas, but not components/headers.
There are a few workarounds for this. I'm testing the best one, so the review will take a bit longer than expected.

